Issue #27, opened 2019-08-12
As a developer, circle CI reports all test reports together, instead of reporting only one failing test suite


DanielSeehausen:
currently, circle ci runs all the tests suites after another and, if any of them fail, it exits and reports red

ideally, as a developer, all the test suites would run and collate their results. even if one fails, it won't break early, and will still run the others.

vdiezel:
I think they do that now because test suites for all services are running in the same step.

PR #39, opened 2019-08-21
28 fix device tests

### LINKED ISSUE #28: As a developer, the tests for /device are both used and pass
currently they are broken, the circle ci config has been commented to not run them at the moment. Fix, refactor, uncomment!

DanielSeehausen:
closes #28 #27 

This is all over the place @vdiezel, so I'll provide comments to draw your attention to the important parts. You may find value in just collapsing all the other files

 - `device/test/device.test.js` can be reviewed as if it is brand new
 - `device/device-utils/` may not be named what we want, or located where we want it - let me know if you have any good ideas for where to put this kind of stuff (e.g. helper methods that act on a model [user/device/process] itself)


.gitignore
```diff
@@ -44,3 +44,4 @@ package-lock.json
 yarn.lock
 .prettier*
 .editorconfig
+*.swp
```
DanielSeehausen: :))

services/device/device-utils/serializeUsers.js
```diff
@@ -0,0 +1,22 @@
+const KEYS_TO_REMOVE = [ 'notifications', '_id' ]
```
DanielSeehausen: this is new file

[serializers are a pretty common practice in APIs](https://api-platform.com/docs/core/serialization/)
and as our api grows, our serializers will. 

if we were being even better citizens, we would keep the database interaction _separate_ from the actual serialization, i.e.:

```javascript
// getDeviceUsers(deviceId)

const device = getDevice(deviceId) // a db call
const users = getDeviceUsers(device) // a db call

const serializedDeviceUsers = serializeDeviceUsers(device, users) // no db call
```

services/device/package.json
```diff
@@ -1,38 +1,18 @@
...
-    ],
-    "*.scss": [
-      "yarn format"
-    ],
-    "*.md": [
-      "yarn format"
-    ]
+  "jest": {
+    "preset": "@shelf/jest-mongodb",
+    "modulePaths": ["<rootDir>/node_modules/"]
```
DanielSeehausen: this was added in all three package.json
there is a way to run the tests outside of the directories where they won't then resolve the correct `node_modules` directory. This forces each one to use its own `node_modules`

services/device/src/device/grantAccess.js
```diff
@@ -1,26 +1,25 @@
...
+    const deviceId = event.pathParameters.id
 
-    const db = await mongo.db();
-    const device = new Device(db);
+    const db = await mongo.db()
+    const device = new Device(db)
 
-    const result = await device.grantAccess(userId, deviceId, data);
-    return response(result);
+    const result = await device.grantAccess(grantingUserId, deviceId, data)
```
DanielSeehausen: minor naming convention to differentiate `data.userId` and `userId`

services/device/test/device.test.js
```diff
@@ -1,1919 +1,950 @@
...
-const addUser = require('../src/device/addUser').main;
-const removeUser = require('../src/device/removeUser').main;
-const modifyAccess = require('../src/device/modifyAccess').main;
-const getDeviceUsers = require('../src/device/getDeviceUsers').main;
-const access = require('../libs/access-rights-lib');
-const mongo = require('../libs/mongodb-lib');
-const uuidv4 = require('uuid/v4');
-
-const DROP_DATABASES = true;
+const uuidv4 = require('uuid/v4')
```
DanielSeehausen: this whole file can be reviewed as if it is new

the crazy amount of imports at the front isn't super. If this grows a lot, we can break these out so each endpoint has its own test file

services/device/test/seeds/device.js
```diff
@@ -0,0 +1,85 @@
+const uuidv4 = require('uuid/v4')
```
DanielSeehausen: `faker` now being used

services/device/test/seeds/stream.js
```diff
@@ -0,0 +1,27 @@
...
+function applyTraits(originalData, traits) {
+  return traits.reduce((newData, trait) => (
+    { ...newData, ...traitDispatch[trait](originalData) }
+  ), originalData)
+}
+
+const generateStream = (data, traits = []) => {
+
+  const streamData = {
+    id: Math.floor(100000 + Math.random() * 900000), // 6 random digits, first never 0
```
DanielSeehausen: is this right for a streamID?

services/process/test/processes.test.js
```diff
@@ -10,7 +10,7 @@ const updateProcess = require('../src/process/update').main
 const cloneProcess = require('../src/process/clone').main
 
 const { generateUser } = require('../../user/test/seeds/user')
-const { getMinimalDevice } = require('../../device/test/seeds/devices')
+const { generateDevice } = require('../../device/test/seeds/device')
```
DanielSeehausen: just naming convention here

services/device/src/device/device.js
```diff
@@ -1,32 +1,38 @@
...
+const { role, DEVICE_ACTIONS, ROLE_ACTIONS, createDefaultGroups, group, RESOURCE } = require('../../libs/access-rights-lib')
 
 class Device {
+
   constructor(db, iotDataLib) {
-    this.db = db;
-    this.iotDataLib = iotDataLib;
+    this.iotDataLib = iotDataLib
+
+    this.usersDB = db.collection('users')
```
DanielSeehausen: this was abstracted here to dry up the interface.
if there is a reason we shouldn't do this (granted, we aren't using both collections in every function, but I don't think that's an issue because we were till saving the reference to the db conn before anyways)

services/device/src/device/device.js
```diff
@@ -124,156 +129,130 @@ class Device {
...
-    const allUsers =  deviceUsers.map(entry => {
-      return {
-        userName: entry.userName,
-        email: entry.email,
-        id: entry.userId
-      };
-    });
-
-    return {statusCode: 200, content: allUsers};
+    const serializedUsers = await this.serializeUsers(device)
```
DanielSeehausen: the original payload should still be intact, only with more keys now (all keys present in user will be serialized, except for `notifications` and `_id`, as they don't seem pertinent to what the receiver might do with a collection of `deviceUsers`

services/device/src/device/device.js
```diff
@@ -284,50 +263,47 @@ class Device {
...
+    const revokingUser = await this.usersDB.findOne({ userId: revokingUserId })
 
-    if (!device){
-      return {statusCode: 404, content: 'Device not found.'};
+    if (!revokingUser) {
+      return {statusCode: 404, content: `User ${revokingUserId} not found.`}
     }
 
-    const deviceUserGroup = device.access.find(group => group.users.includes(userId));
+    const revokingUserDeviceGroup = device.access.find(group => group.users.includes(revokingUserId))
```
DanielSeehausen: might be missing a case where the revokingUser isn't found at all on the device

services/device/src/device/device.js
```diff
@@ -336,116 +312,113 @@ class Device {
...
-      $pull: {users: user.userId, 'access.$.users': user.userId}
-    });
+    const filter = { deviceId, 'access.name': data.groupName }
+    const result = await this.devicesDB.updateOne(filter,
+      { $pull: { users: receivingUser.userId, 'access.$.users': receivingUser.userId} }
+    )
 
     if (result.modifiedCount !== 1){
-      return {statusCode: 400, content: 'Could not remove user from device.'};
+      return {statusCode: 500, content: `Could not remove user ${receivingUser.userId} from device for an uncaught reason.`}
```
DanielSeehausen: note 400 => 500
if we hit it, we done goofed

services/device/src/device/device.js
```diff
@@ -457,178 +430,167 @@ class Device {
...
-        $push: {streamNameAliases: aliasEntry}
-      });
-    }
-    else {
-      result = await devices.updateOne({ deviceId, 'streamNameAliases.streamId': data.streamId }, {
-        $set: {
-          'streamNameAliases.$.alias' : data.alias
-        }});
+    if (existingAlias) {
+      if (existingAlias.alias === data.alias) // do nothing if the alias string
```
DanielSeehausen: oops meant to remove that comment :/

services/device/src/device/device.js
```diff
@@ -124,156 +129,130 @@ class Device {
...
-    return {statusCode: 200, content: result.deletedCount};
+    return {statusCode: 200, content: result.deletedCount}
   }
 
   async getUsers(userId, deviceId){
-    const devices = this.db.collection('devices');
-    const users = this.db.collection('users');
-
-    const device = await devices.findOne({ deviceId });
+    const device = await this.devicesDB.findOne({ deviceId })
```
DanielSeehausen: this one probably changed the most, where users are no longer derived from the `users` array, but instead the access groups

services/device/device-utils/serializeUsers.js
```diff
@@ -0,0 +1,22 @@
...
+  for (let group of device.access) {
+
+    for (let userId of group.users) {
+      const user = await this.usersDB.findOne({ userId })
+      KEYS_TO_REMOVE.forEach(key => delete user[key])
+
+      serializedUsers.push({
+        ...user,
+        accessGroup: group.name,
+        id: userId // duplicate of userId, but as long as we are using both throughout the code base we should be returning both keys to avoid ambiguity. TODO: deprecate usage of both userId and id in user payloads
```
vdiezel: As of my knowledge, we aren't  using the `id` or (`_id`) anywhere in the code! In case we do, it can be considered a mistake and should be fixed. Maybe you can pin point where we are using the `id` that is beeing assigned by MongoDB?

services/device/libs/access-rights-lib.js
```diff
@@ -112,18 +112,18 @@ function createProcessGroup(name){
   return group;
 }
 
-exports.createDefaultGroups = function (type, ownerId = undefined){
+exports.createDefaultGroups = function (type, ownerId = undefined) {
 
   let groups = [];
 
-  if (type === RESOURCE.PROCESS){
+  if (type === RESOURCE.PROCESS) {
```
vdiezel: OCD strikes again

services/device/src/device/create.js
```diff
@@ -7,9 +8,7 @@ exports.main = async function main(event, context) {
   try {
     context.callbackWaitsForEmptyEventLoop = false;
 
-    const userId = event.requestContext.identity.cognitoAuthenticationProvider.match(
-      /([^CognitoSignIn:]+$)/g
-    )[0];
+    const userId = parseCognitoIdFromEvent(event)
```
vdiezel: was about time to do this!

services/device/test/seeds/stream.js
vdiezel: That's pretty much on point.

services/device/test/utils/mockIotDataLib.js
```diff
@@ -0,0 +1,9 @@
+// janky, but need to silence this and mock its usage to get this test passing for setCalibration. Should find a longer term/testable solution for this at some point... ds 08/19
+//
+process.env.AWS_IOT_ENDPOINT = 'test iot endpoint'
+
+const iotDataLib = require('../../libs/iotData-lib')
+
+iotDataLib.iotData.publish = jest.fn(() => ({
```
vdiezel: :1st_place_medal:

services/device/test/device.test.js
vdiezel: Looks good!

services/device/device-utils/serializeUsers.js
DanielSeehausen: i removed it now
I put this in here because it looked like the original response for `getUsers` (which I think is the method that this helper is for) included user payloads with an id key. 

If you are confident that it's not being used in the code base its totally ok to remove. I just wanted to be careful not to provide a breaking change.

services/device/libs/access-rights-lib.js
DanielSeehausen: 👁

PR #49, opened 2019-09-03
added entryPhaseId when process is created

vdiezel:
backend changes for the dispatching of process changes to the device


services/device/src/command/command.js
```diff
@@ -1,24 +1,6 @@
 const { DEVICE_ACTIONS, PROCESS_ACTIONS, group, RESOURCE } = require('../../libs/access-rights-lib')
 const ObjectID = require('mongodb').ObjectID
 
-//////////////////////////////////////////////////////////////////////////
```
DanielSeehausen: 😀

services/device/src/command/command.js
```diff
@@ -59,13 +41,15 @@ class Command {
       return { statusCode: 403, content: 'No Authorization for this process operation: Insufficient permission' }
     }
 
+    const command = data.update ? 'update' : 'start'
```
DanielSeehausen: this works for now np and good to avoid YAGNI hell, but understanding the firmware side is splitting the commands up because their uses may change, this might be broken out later. who knows!

PR #50, opened 2019-09-03
removed last initialize usage from streamDefaults

### LINKED ISSUE #37: remove usage of `initialize` in event payload handling in the api
~to be completed after #20 , which provides test coverage that should make this a lot simpler~

This is ready to go!

perviously, we needed to use the `initialize` key throughout the code base. It is no longer required, and we are ready to deprecate! This will touch a lot of things lightly, most likely, and require some mongo query updates, most likely.

vdiezel:
closes #37


services/device/src/device/device.js
```diff
@@ -535,7 +531,7 @@ class Device {
 
   async setStreamNameAlias(userId, deviceId, data){
     const streamIdMissing = !data.streamId
-    const streamIdInvalid = !Number.isInteger(data.streamId)
+    const streamIdInvalid = !Number.isInteger(parseInt(data.streamId))
```
vdiezel: streamId is sent a s a string, needed to parse here

I didn't implement an equivalent check for the calibration and the stream defaults. What we want to do in the future is, instead of checking every single input for the right format, using a package like Joi and validate the complete request. This should make the logic more clean and modularized!

services/device/src/device/device.js
```diff
@@ -488,25 +488,21 @@ class Device {
...
-      result = await this.devicesDB.updateOne({ deviceId }, {
-        $push: {streamDefaults: defaultEntry}
-      })
-    }
-    else {
-      result = await this.devicesDB.updateOne({ deviceId, 'streamDefaults.streamId': data.streamId}, {
-        $set: {
-          'streamDefaults.$.value' : data.value
-        }})
+    if (existingDefault) {
```
DanielSeehausen: these `ifs` can be combined -- i think the nesting isn't doing anything for us here

services/device/test/seeds/event-generators/setStreamDefaultEvent.js
```diff
@@ -3,7 +3,6 @@ const setStreamDefaultEvent = (device, stream, requestingUserId) => ({
     id: device.deviceId,
   },
   body: JSON.stringify({
-    initialize: true,
```
DanielSeehausen: 🕵 nice laser eyes viktor

DanielSeehausen:
🚢 it!

services/device/src/device/device.js
DanielSeehausen: im 100% game for this JOI integration.
Will slack about it

PR #54, opened 2019-09-05
added new function to get a process data summary

vdiezel:
added a new function that fetches a process data summary (3000 datapoints of every stream of the process) to serve data quickly to the client on the initial process loading 

belongs to REACTor#119


services/process/src/dataPoints/getDataSummary.js
```diff
@@ -0,0 +1,24 @@
+const mongo = require('../../libs/mongodb-lib');
+const response = require('../../libs/response-lib').response;
+const DataPoints = require('./dataPoints');
+
+exports.main = async function main(event, context) {
+  try {
+    context.callbackWaitsForEmptyEventLoop = false;
+
+    const userId = event.requestContext.identity.cognitoAuthenticationProvider.match(
```
DanielSeehausen: [we have a helper for this now (you will have to copy the lib file into process libs though)](https://github.com/ospin-web-dev/hambda/blob/master/services/device/libs/request-lib.js)

once the layering thing is done, they will be shared

services/process/src/dataPoints/dataPoints.js
```diff
@@ -74,6 +74,68 @@ class DataPoints {
...
+        })
+      }
+    }
+
+    // reduce to maximal numberOfPoints by skipping elements
+
+    const NUMBER_OF_POINTS = 3000
+    const res = { data: [] };
+
+    streamIds.forEach(streamId => {
```
DanielSeehausen: couple of really cheap ways to clean this up:

- create the `res` wrapper after the iteration (see below)
- change the iteration to a `const data = map(etc...)`
- either fetch all of the streams first into an array, or assign the stream at the top of the iteration (`const stream = normalizedData[streamId]`)
- update all calls to the `normalizedData[x]` to just the stream
- either wrap the res now, or just save a line
  `{ statusCode: 200, content: { data } }`

PR #81, opened 2019-09-26
79 public get user endpoint

### LINKED ISSUE #80: create a public endpoint for GET user
related to ospin-web-dev/reactor#112
returns only the name and id

DanielSeehausen:
I split the commits up in a way that would make even @torvalds happy, (and provided minimal commit messages), so I'd recommend reviewing them in order instead of just looking at all the file changes at once @vdiezel @petros751 

closes #80 

@vdiezel I am unsure what is left to do on the sls side of things. Would you expect it to work as is or is there something left to be set up.


services/user/src/serializers/public/user.js
```diff
@@ -0,0 +1,8 @@
+const pick = require('utils').pick
+
+
+const ALLOWED_ATTRIBUTES = [ 'id', 'userName' ]
+
+module.exports = async (user) => (
+  pick(...ALLOWED_ATTRIBUTES)(user)
```
DanielSeehausen: dank

vdiezel: dank²

PR #86, opened 2019-09-30
Refac data storage

### LINKED ISSUE #85: change data chunks saving architecture to use a single collection for all processes

vdiezel:
Changes the way we store data for the processes - we used to save data chunks for every process in a different collection- but due to resctrictions we now store all data chunks in a single collection and use indexing to query the data quicker!

closes #85 

_REQUIRES_ a `npm i`  in the process service after merging and pulling because a package (csv-writer) was missing!


lib/utils.js
```diff
@@ -1,12 +1,8 @@
-exports.dataPointsCollectionName = function(processID) {
-  return `dataPoints_${processID}`;
-};
+const region = process.env.OSPIN_AWS_REGION;
```
vdiezel: cleaning up the file

services/process/package.json
```diff
@@ -23,6 +23,7 @@
   },
   "dependencies": {
     "aws-sdk": "^2.224.1",
+    "csv-writer": "^1.5.0",
```
vdiezel: got lost somehow

PR #90, opened 2019-10-03
Dry user device process model

### LINKED ISSUE #69: DRYing up the user, process and device model code
We have a lot of repition, especially when we check for access rights, in the service model files. This should be DRYed up.

### LINKED ISSUE #88: Provide more test coverage for creating, cloning and deleting processes
to guarantee that the reference to the process is set/removed correctly to the devices und users `processes` field

### LINKED ISSUE #95: incorporate middy with example usage of auth super user on user/get
see title

vdiezel:
A big one!

- refactored response API
- abstracted the validation within the business logic into a `validator` function
- implemented JOI on the user service
- provided more test coverage on the user service

closes #69 
progresses #67 by a big deal
closes #88 
closes #95 

massive change. APIv2!
we tested this over weeks in dev and in function isolation. It's way too big of a PR, and that is my fault, but its not as extreme as it seems as we were able to split up the testing. Will sit in dev for a couple more weeks before going to prod!


lib/response-lib.js
```diff
@@ -1,24 +1,20 @@
...
-    },
-    body: JSON.stringify(result.content)
-  };
-};
+    }
+  }
 
+  response.body = JSON.stringify(content)
+  return response
+};
```
vdiezel: predefined responses that are used in lambda handlers and the `validator`

lib/user-joi-schemas.js
```diff
@@ -0,0 +1,78 @@
+const Joi = require('@hapi/joi')
```
vdiezel: joi schemas for user service

lib/userlog-joi-schemas.js
```diff
@@ -0,0 +1,35 @@
+const Joi = require('@hapi/joi')
```
vdiezel: schemas for user logs (endpoints not in use yet)

lib/validation-lib.js
```diff
@@ -1,13 +1,56 @@
...
 
-const newProcessSchema  = {
-  name: Joi.string().min(1).max(45).required(),
-  deviceName: Joi.string().required(),
-  description: Joi.object(),
-  comment: Joi.string().allow(''),
-  deviceDescription: Joi.array(),
-  cloneHistory: Joi.array(),
-  entryPhaseId: Joi.number().integer().required()
+exports.validator = function validator(validators) {
```
vdiezel: validator function that takes in an array of validator functions which return ` { invalid : true , response: ... }` when a validation fails. Using `some` here because it will shoort circuit the call on the first time we have a failed validation. Further down are the different validation functions I used so far

vdiezel: I provided a bit of test coverage for this one, however I didn't know where the best location for those general `lib`-tests is. We have to talk about this @DanielSeehausen . Currently I just dumped the test into `user`-service

services/device/src/command/command.js
```diff
@@ -35,29 +37,11 @@ class Command {
...
     const process = await processPromise
 
-    if (!device) {
-      return { statusCode: 404, content: 'Device not found.' }
-    }
-
-    const deviceUserGroup = device.access.find(group => group.users.includes(userId))
-
-    if (!deviceUserGroup) return { statusCode: 403, content: 'No Authorization for this device operation: No group found.' }
+    const processErrorResponse = validator([ () => canFind(process, `Process ${processId}`), () => hasObjectAccess(process, userId, RESOURCE.PROCESS, PROCESS_ACTIONS.STOP) ])
```
vdiezel: typical usage of the `validator`

lib/validation-lib.js
```diff
@@ -1,13 +1,56 @@
...
+exports.hasObjectAccess = function hasObjectAccess(resource, userId, resourceType, action) {
+  const resourceUserGroup = resource.access.find(group => group.users.includes(userId))
+  if (!resourceUserGroup)
+    return { invalid: true, response: noAuthorization('Group not found')}
+  if (group(resourceType, resourceUserGroup.rights).cant(action))
+    return { invalid: true, response: noAuthorization('Insufficient permission')}
+  return { invalid: false }
+}
+
+exports.isMyResource = function isMyResource(requestingUserId, targetUserId) {
```
DanielSeehausen: love it

lib/validation-lib.js
DanielSeehausen: love this pattern!
`validate` might be a better name, because when invoked, it performs validation

lib/validation-lib.js
```diff
@@ -1,13 +1,56 @@
...
+    return { invalid: true, response: noAuthorization(`This resource can only be accessed by user ${targetUserId}`)}
+  return { invalid: false }
 }
 
-exports.newProcessSchema = newProcessSchema
+exports.isNotPresentYet = function isNotPresentYet(resources, id, label) {
+  if (resources.includes(id))
+    return { invalid: true, response: badRequest(`The resource ${id} is already present in ${label}`)}
+  return { invalid: false }
+}
```
DanielSeehausen: we aren't making use of `label` here except in the error message.

is this suppose to assert whether an array of objects includes an object that has a positive match on 
`o[label] === id`?

lib/validation-lib.js
```diff
@@ -1,13 +1,56 @@
...
+    return { invalid: true, response: noAuthorization('Group not found')}
+  if (group(resourceType, resourceUserGroup.rights).cant(action))
+    return { invalid: true, response: noAuthorization('Insufficient permission')}
+  return { invalid: false }
+}
+
+exports.isMyResource = function isMyResource(requestingUserId, targetUserId) {
+  if (requestingUserId !== targetUserId)
+    return { invalid: true, response: noAuthorization(`This resource can only be accessed by user ${targetUserId}`)}
+  return { invalid: false }
```
DanielSeehausen: I love it

because this isn't asked at the resource level (we don't pass a User or Device object in against a requestingID), this might be more appropriate as a generic `idsEqual(a, b)` or something.

services/device/src/command/command.js
DanielSeehausen: we can combine these with the structure you provided in `validator` I think
`processErrorResponse` and `deviceErrorResponse`

```javascript
const error = validator([
  () => canFind(process, `Process ${processId`),
  () => hasObjectAccess(process, userId, RESOURCE.PROCESS, PROCESS_ACTIONS.STOP),
  () => canFind(device, `Device ${data.deviceId}`),
  () => hasObjectAccess(device, userId, RESOURCE.DEVICE, DEVICE_ACTIONS.OPERATE) 
])

**Comment by DanielSeehausen** [comment]
this is still in process
`user` is finished. 
I will deploy just the user functions and assert they are working before moving on to process and device

**Comment by DanielSeehausen** [comment]
this PR is crazy big, but its because we can test the parts in isolation, so forgive us whoever reads this message in 2030

File: lib/response-lib.js [hunk 9]
```diff
@@ -1,24 +1,16 @@
...
-  };
-};
+exports.response = ({ statusCode, body }) => ({
+  statusCode,
+  headers: {
+    'Access-Control-Allow-Origin': '*',
+    'Access-Control-Allow-Credentials': true
+  },
+  body: JSON.stringify(body)
+})
```
DanielSeehausen: once the PR is finished, we should be able to get rid of this entirely and ensure middy does the response

services/device/src/command/command.js
DanielSeehausen: this is outdated now. see `assertValidators`

services/device/src/iotTrigger/trigger.js.orig
```diff
@@ -0,0 +1,88 @@
+<<<<<<< HEAD
+const serializeStreamsData = require('../serializers/serializeStreamsData')
+=======
+const { success } = require('response-lib')
+>>>>>>> WIP: refac device
```
DanielSeehausen: oops

services/process/src/dataPoints/requestDownloadStreamData.js
```diff
@@ -1,14 +1,13 @@
 const mongo = require('mongodb-lib');
```
vdiezel: files in process/src/dataPoints still have the old API

services/user/src/cognito/CognitoAPI.js
```diff
@@ -0,0 +1,34 @@
+const AWS = require('aws-sdk')
+
+const USER_POOL_ID = process.env.AWS_COGNITO_USERPOOL_ID
+const COGNITO_CLIENT = new AWS.CognitoIdentityServiceProvider({ region: process.env.OSPIN_AWS_REGION })
+
+
+const genReq = (data) => ({
```
vdiezel: I would just use `generateRequest` here - I think we can afford the bytes ;)

PR #94, opened 2019-10-10
236 reactor graph display bug

DanielSeehausen:
closes ospin-web-dev/reactor#236

easiest to review if you look at the two commits separately!


services/device/src/serializers/serializeStreamsData.js
```diff
@@ -0,0 +1,19 @@
...
+  Object.entries(streamsData).map(([ streamId, [ times, values ] ]) => {
+    const { newA: sortedTimes, newB: sortedValues } = sortInParallel(times, values)
+    return {
+      id: streamId,
+      start: sortedTimes[0],
+      times: sortedTimes,
+      values: sortedValues
+    }
+  })
+)
```
vdiezel: :+1:

services/device/src/iotTrigger/trigger.js
```diff
@@ -12,16 +14,10 @@ class Trigger {
...
-        start: values[0][0],
-        times: values[0],
-        values: values[1]
-      };
-    });
+    // streamsData = { streamId: [ [times], [values] ] }
+    const serializedStreams = serializeStreamsData(streamsData)
 
-    for (let stream of startingTimes){
+    for (let stream of serializedStreams){
```
vdiezel: `let stream` can go to a `const stream` - JavaScript creates a new variable on every iteration instead of reusing the old one - that is why it works! 

https://stackoverflow.com/questions/41067790/why-does-const-work-in-some-for-loops-in-javascript

test/lib/sorts.test.js
```diff
@@ -0,0 +1,60 @@
...
+      expect(newB).not.toBe(shortArrB)
+    })
+
+    test('warns in console', () => {
+      const { newA, newB } = sortInParallel(arrA, shortArrB)
+
+      expect(console.warn).toHaveBeenCalled()
+    })
+  })
+})
```
vdiezel: very nice!

lib/sorts.js
```diff
@@ -0,0 +1,30 @@
...
+
+function warnStack(msg) {
+  const stack = new Error().stack
+  console.warn(msg, stack)
+}
+
+exports.sortInParallel = (arrA, arrB, sort = sortByAsc) => {
+  if (arrA.length !== arrB.length) {
+    console.warn('Arrays must be equal length')
+    return { newA: arrA.slice(), newB: arrB.slice() }
```
vdiezel: Maybe use `[ ... arrA ]` instead of `slice` because its more clear about whats happening (copying, and not taking a slice)

lib/sorts.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  const indices = arrA.map((_, i) => i)
+
+  indices.sort((ai, bi) => (
+    (arrA[ai] < arrA[bi]) ? -1 : (arrA[ai] > arrA[bi]) ? 1 : 0
```
vdiezel: maybe extract that sort algorithm and give it a meaningful name so we now whats going on here, just like `sortByAsc`

lib/sorts.js
DanielSeehausen: good point - this is a relic of the past (before the spread operator, slice was the de-facto array copier)
I guess I still haven't kicked the habit yet!

lib/sorts.js
DanielSeehausen: this was an oversight on my part. I wanted to use the sortByAsc function

PR #104, opened 2019-11-06
added fallback for a runningProcessId of "null" (String) to be null (…

### LINKED ISSUE #103: Provide a fallback for a runningProcessId of "null" as a string to be the primitive value of null
The firmware sends up (sometimes) the processId for state updates on `:env/state/:deviceId` as the string "null" now, when it should not be present in the payload at all when it is supposed to be set null in the DB. Provide a fallback to guarantee the processId is the value null, never the string "null"

vdiezel:
…value)

closes #103


vdiezel:
Integrated in the DRY-..... branch instead.

PR #110, opened 2019-11-12
Add notifications for user

DanielSeehausen:
adds `patchAllNotifications` for the new changelog updates
has test coverage

please let me know if the serverless file is right @vdiezel!


services/user/src/user/user.js
```diff
@@ -198,6 +200,29 @@ class User {
...
+      id: uuidv4(),
+      topic,
+      type: 'ephemeral',
+      from: {
+        type: sourceType,
+        name: sourceName,
+        originId: requestingUserId
+      },
+      data: {
+        message: 'There are unseen updates to the web application.'
```
vdiezel: Dont we want to actually integrate which change happened here, too? Maybe adding te titleof the corresponding PR?

DanielSeehausen: that is a good idea. ill make the change

services/user/src/user/user.js
```diff
@@ -198,6 +200,29 @@ class User {
...
+      }
+    }
+
+    const result = await this.usersDB.updateMany(
+      { 'notifications.topic': { '$ne': topic } },
```
vdiezel: OK - now I fully understand your approach. You want to use the notification like a flag and diplay a generic text on the frontend. In this case I'd prefer just setting a flag on the user and using the existing `updateUser` endpoint. 

Using the notifications makes sense when we actually add custom information about the update (like described in my other comment) and then add them to all users (instead of querying the users who do not have the `flag` notification on them!

DanielSeehausen:
going to make these changes now

services/user/src/user/user.js
DanielSeehausen: function now accepts an optional 'message' key in the body

services/user/src/user/user.js
DanielSeehausen: updated with the message key

DanielSeehausen: and now one user can have many notifications

PR #111, opened 2019-11-13
updates several minor things on patchallnotificat

DanielSeehausen:
- update expected payload to allow an optional message as a string
- update mongo insert to add multiple update notifications to users,
instead of only one
- update tests
- update notification type to persistent

[see pr comments here](https://github.com/ospin-web-dev/hambda/pull/110)


services/user/src/user/user.js
```diff
@@ -201,24 +201,22 @@ class User {
     return result.modifiedCount
   }
 
-  async patchAllNotifications({ requestingUserId, sourceName, sourceType, topic }) {
-    const changelogNotification = {
+  async patchAllNotifications({ requestingUserId, sourceName, sourceType, topic, message }) {
+    const newNotification = {
```
DanielSeehausen: name changed because it is not generic. only thing missing is persistent/ephemeral option, but I figure YAGNI for now!

services/user/test/seeds/user.js
```diff
@@ -9,7 +9,7 @@ const traitDispatch = {
     notifications: [{
       id: uuidv4(),
       topic: 'changelog',
-      type: 'ephemeral',
+      type: 'persistent',
```
DanielSeehausen: because this is specifically a 'changelog' notification (named attribute on the user seed)

services/user/test/user.test.js
```diff
@@ -468,24 +468,40 @@ describe('User Class', () => {
...
+
+      expect(newNot.topic).toEqual(topic)
+      expect(newNot.from).toEqual(expect.objectContaining({
+        type: sourceType,
+        name: sourceName,
+        originId: requestingUserId
+      }))
+      expect(newNot.data).toEqual(expect.objectContaining({
+        message
+      }))
```
DanielSeehausen: now testing structure too!
Should blow up quicker if someone starts changing the api. Blow up in a good way, that is :fire:

services/user/test/user.test.js
```diff
@@ -468,24 +468,40 @@ describe('User Class', () => {
...
+      const notificationIds = userB.notifications.map(not => not.id)
+      expect(notificationIds).toEqual([ existingNot.id, newNot.id ])
+
+      expect(newNot.topic).toEqual(topic)
+      expect(newNot.from).toEqual(expect.objectContaining({
+        type: sourceType,
+        name: sourceName,
+        originId: requestingUserId
+      }))
+      expect(newNot.data).toEqual(expect.objectContaining({
```
vdiezel: :100:

services/user/src/user/patchAllNotifications.js
```diff
@@ -24,7 +24,8 @@ async function patchAllNotifications(event) {
     requestingUserId: event.requestingUserId,
```
vdiezel: there will be no requestingUserId set for this one - I can provide you something that identifies CircleCI as the origin Id, or maybe the commit hash so we now where where this notification cam from

vdiezel: als we don't need the `setRequestingUserId` in the middlewares of the lambda `main` handler

PR #115, opened 2019-11-14
GitHub api call

O-Susanoo:


services/user/src/user/user.js
```diff
@@ -1,19 +1,22 @@
...
 
 class User {
 
   constructor(db) {
-    this.db = db
-    this.usersDB = db.collection('users')
-    this.devicesDB = db.collection('devices')
-    this.processesDB = db.collection('processes')
+    if(db){
+      this.db = db
```
DanielSeehausen: this is a mistake of mine. we shouldn't need to keep track of the `db` variable here in on the instance. 
[here in delete](https://github.com/ospin-web-dev/hambda/pull/115/files#diff-7e80efd8982a2879fb5f6d57c1faed20R60-R61) I accidentally left two calls to the `this.db`. They should just be `this.processesDB` and `this.devicesDB`. Then we can remove `this.db` all together. Would you mind fixing, since this code touches it?
Would you be able to

services/user/src/user/user.js
```diff
@@ -223,6 +226,25 @@ class User {
...
     return result.modifiedCount
   }
+
+  async getChangelog(){
+    const PR_COUNT = 10
+    const response = await fetch('https://api.github.com/repos/ospin-web-dev/REACTor/pulls?state=closed', {
+      headers: {
+        'Authorization': `token ${process.env.GITHUB_ACCESS_TOKEN}`
+      }
+    });
```
DanielSeehausen: not a request to change stuff, just perhaps important to note:

right now we have no error handling condition for this. One side effect of no longer using the API directly is that we lose a lot of the nice work the API automatically does (error handling/response codes/etc.)

A way we can avoid all this work is just to ferry the response from github directly back to our front end, then let the front end deal with it. That way we keep the work their massive dev team as already done.

don't need to change anything now or for this PR, but just for future reference we should do some 'defensive programming' here and assert that the response we got back was a successful response with valid json. (so that if it isn't, we can meaningfully communicate back to the requestor.)

PR #117, opened 2019-11-14
changed patchAllNotifications to fit event invocation from CircleCI

vdiezel:
Modifies the patchAllNotifications to work with circleCI/ aws-sdk lambda direct invocation pattern


services/user/src/joi/user.js
```diff
@@ -41,11 +41,11 @@ const deleteNotification = assertSchema(Joi.object({
 
 
 const patchAllNotifications = assertSchema(Joi.object({
-  requestingUserId: Joi.string().required().pattern(UUIDV4_REGEX),
+  originId: Joi.string().required(),
```
DanielSeehausen: 👍

services/user/src/joi/user.js
```diff
@@ -41,11 +41,11 @@ const deleteNotification = assertSchema(Joi.object({
...
   sourceName: Joi.string().valid('circleCI').required(),
   sourceType: Joi.string().valid('mergeHook').required(),
   topic: Joi.string().valid('changelog').required(),
-  message: Joi.string()
+  message: Joi.string().required().allow('')
```
DanielSeehausen: slacked you re: this!

services/user/src/user/patchAllNotifications.js
```diff
@@ -20,8 +20,9 @@ const { logger } = require('envDependentLogger')
 const User = require('./user')
 
 async function patchAllNotifications(event) {
+
   const requestData = {
-    requestingUserId: event.requestingUserId,
+    originId: event.body.originId,
```
DanielSeehausen: this now means we can no longer call this from the client/nexus and have the userId automatically injected. I don't see any consequences, but worth noting that it is a departure from how most of our stuff works. 

as long as this line is still actually forcing authorization I am a happy camper: `authorizer: aws_iam`

services/user/test/seeds/event-generators/patchAllNotificationsEvent.js
```diff
@@ -1,14 +1,6 @@
-const patchAllNotificationsEvent = (requestingUserId, sourceName, sourceType, topic, message) => ({
-  headers: {
-    'Content-Type': 'application/json',
-    'content-type': 'application/json'
-  },
```
DanielSeehausen: why did the headers get removed? were they breaking something?

services/user/src/user/patchAllNotifications.js
vdiezel: yes it does still force authorization - in fact, any cognito user from our application (probably not what we want) or any user who is registered on our AWS account and has the proper access rights can invoke this function. 
I had the same thoughts here about changing the way we interact with the backend and the only way to bypass that is to fake a HTTP request and pass it as an event from the CircleCI lambda invocation.

services/user/test/seeds/event-generators/patchAllNotificationsEvent.js
vdiezel: I removed them because it's not an http request anymore.

services/user/serverless.yml
```diff
@@ -348,12 +348,12 @@ functions:
...
-          path: /user/notifications
-          method: patch
-          cors: true
-          authorizer: aws_iam
+    # events:
+    #   - http:
+    #       path: /user/notifications
+    #       method: patch
+    #       cors: true
+    #       authorizer: aws_iam
```
ospin-webapp: made the patchAllNotifications not accessable via HTTP fro now because we invoke the function directly via lambda. Until we haven't figured out how to made this endpoint flexible for both kind of requests HTTP and direct invocation via aws-cli lambda) - otherwise this endpoint will break when invoked via HTTP

PR #121, opened 2019-11-14
fix devicelogs endpoints

### LINKED ISSUE #122: devicelogs api throws errors and causes the webapp to crash in `log` tab

vdiezel:
fixes device log bug by importing the success function from the correct library and adjust passing down data using the new `dank` pattern by @DanielSeehausen 
closes #122


services/device/src/log/create.js
```diff
@@ -13,24 +13,22 @@ const {
   jsonBodyParser
 } = require('middy/middlewares')
 
-const { success } = require('success-lib')
+const { success } = require('response-lib')
```
DanielSeehausen: 👀 my fault!

PR #132, opened 2019-11-18
Process joi implementation

### LINKED ISSUE #97: add JOI request schemas for /process

vdiezel:
closes #97 

Adds Joi implementation to the user service and test coverage for managing access to processes
Provided comments where important, but in general I applied the same patterns as in the user service

98% line changes are due to added tests.


services/process/src/process/process.js
```diff
@@ -193,7 +201,7 @@ class Process {
 
     const clonedProcess = await this.processesDB.insertOne(process);
 
-    await this.devicesDB.updateOne({ id: data.deviceId }, {
+    await this.devicesDB.updateOne({ id: process.deviceId }, {
```
vdiezel: deviceId doesn't need to be passed down to the API because it is present on the process

services/process/src/process/process.js
```diff
@@ -212,38 +220,32 @@ class Process {
...
 
-    const userIds = process.users;
-
-    const processUsers = await this.usersDB.find({ id: {$in : userIds}}).toArray();
-
-    const allUsers =  processUsers.map(entry => {
-      return {
-        username: entry.userName,
-        email: entry.email,
-        id: entry.userId
```
vdiezel: should shouldn't have worked in the first place, userId is not present on users anymore!

services/process/src/process/process.js
```diff
@@ -212,38 +220,32 @@ class Process {
...
     assertMany([
-      () => canFind(process, `Process ${processId}`), () => hasObjectAccess(process, requestingUserId, RESOURCE.PROCESS, PROCESS_ACTIONS.SHARE),
-      () => canFind(user, 'User')
+      () => canFind(process, `Process ${processId}`),
+      () => hasObjectAccess(process, requestingUserId, RESOURCE.PROCESS, PROCESS_ACTIONS.SHARE),
+      () => canFind(user, `User ${data.email}`)
     ])
 
-    const newUserId = user.id;
+    const group = process.access.find(group => group.name === data.groupName)
```
vdiezel: also assert now if the group actually exists instead of silent failure

services/process/src/serializers/serializeUsers.js
```diff
@@ -0,0 +1,21 @@
+const KEYS_TO_REMOVE = [ 'notifications', '_id' ]
```
vdiezel: we have the same function now in two services, we probably should abstract it to top level libs

services/process/test/process.test.js
```diff
@@ -541,5 +1525,26 @@ describe('Process Class', () => {
...
+      const badIds = [ 123, 'yvan eht nioj', undefined ]
+
+      for (const badId of badIds) {
+        const { res, body } = await fireGetPublicProcessEvent(badId)
+        expect(res.statusCode).toBe(422)
+        assertContainsAll(body, ['requestedProcessId'])
+      }
+    })
   })
 })
```
vdiezel: Nothin groundbreaking in the test file - I created a new issue to actually break test files up into smaller files

services/process/src/joi/process.js
```diff
@@ -0,0 +1,99 @@
+const Joi = require('@hapi/joi')
+const { assertSchema } = require('validation-lib')
+
+const UUIDV4_REGEX = /[a-f0-9]{8}(-[a-f0-9]{4}){3}-[a-f0-9]{12}/
```
DanielSeehausen: this is repeated across the board. might be a good opportunity to throw it in `lib/test-helpers/test-helper`.

services/process/src/process/process.js
DanielSeehausen: good catch - much better with process here being populated from the DB instead of some payload a user sends...

services/process/src/process/process.js
DanielSeehausen: 🕵

services/process/src/serializers/serializeUsers.js
DanielSeehausen: good point. I'm wondering if the process and device schemas change over time, would this end up being more work? The side effect I see of using the same method is that we then say "device and process schemas need to always have the same way of accessing users". If they were to change, our serializer would break :(. 
 
TL;DR: I can see a world where it makes sense to keep these separate. All the same, we haven't yet violated WET (the everything twice WET, not the cheeky 'every time' WET) :)

services/process/test/process.test.js
```diff
@@ -118,59 +159,157 @@ describe('Process Class', () => {
...
+      const badArgs = [ 123, true, [], '*'.repeat(251) ]
+      const { user, device, process } = await seedUserDeviceProcess()
+
+      for (const badArg of badArgs){
+        process.comment = badArg
+        const { res, body } = await fireCreateProcessEvent(process, user)
+        expect(res.statusCode).toBe(422)
+        assertContainsAll(body, ['data.comment'])
+      }
+    })
```
DanielSeehausen: good opportunity for another `describe('when a user provides invalid request data...or something...` block. 

nothing worth doing now though, just something that helps future devs easily organize their code in instances where they realize 'oh shoot I need to add another test for bad data'

services/process/test/seeds/event-generators/cloneProcessEvent.js
```diff
@@ -8,8 +9,7 @@ const cloneProcessEvent = (processId, cloneData, requestingUserId) => ({
   },
   body: JSON.stringify({
     processName: cloneData.processName,
-    processId: cloneData.processId,
-    deviceId: cloneData.deviceId
```
DanielSeehausen: good 👁 removing this stuff from everywhere

services/process/src/joi/process.js
vdiezel: If I understand you correct, yo would like to abstract the `Joi.string().required().pattern()` call ?

vdiezel:
Regarding your concerns: Nexus talks (mostly) to API Gateway, the events are passed down from API Gateway to Lambda, not from nexus. and I do think that we are already quite close in mocking those events for the relevant part of the data. I'm not completely sure what you mean by incompatible. That the event mocks we are creating do not represent sufficiently real events coming in?

For tetsing we could do something like this:
We could write a testing suite that uses the aws-cli to run so called `dry` requests on our Lambda functions, but it would invoke them directly. Because we do not want to modify the lambda handlers, we would end up writing http fake events again. This would give us the possibility to test functions in a deployed state. If you want add nexus into it, we could do the following: mock nexus functions to call aws-cli in dry mode, passing down the args the nexus funcions expects, create an HTTP event from it (like we do on hambda) and pass it directly to lambda and see if we get a proper response. Dry runs do not execute any of our business logic which should be covered by our unit tests.

PR #133, opened 2019-11-19
Fix : close all mongodb opened handles to fix jest test

### LINKED ISSUE #140: fix open handlers in test suite

O-Susanoo:
closes #140


lib/mongodb-lib.js
```diff
@@ -1,10 +1,11 @@
 const MongoClient = require('mongodb').MongoClient;
-
+let conn = null;
```
DanielSeehausen: don't forget a space or double space between deps and code please!

lib/mongodb-lib.js
```diff
@@ -27,3 +28,6 @@ async function getProdDB() {
 exports.db = async function db() {
   return await process.env.NODE_ENV === 'test' ? getTestDB() : getProdDB()
 };
+exports.conn = ()=>{
+  return conn
+};
```
DanielSeehausen: might need to change your newline setting on that computer as well

O-Susanoo: Thank you for the update , i'm now on eslint theoretically i should not repeat those mistakes

PR #142, opened 2019-11-22
Issue 105 handle iot publish

### LINKED ISSUE #105: Provide proper server response when IoT command publishes fail

O-Susanoo:
Closes #105


lib/IotDataAPI.js
```diff
@@ -1,12 +1,29 @@
...
+    if (result.$response.httpResponse.statusCode === 401) {
+      return new Promise(() => iotUnauthorized(result.$response.error))
+    }
+    if (result.$response.httpResponse.statusCode === 405) {
+      return new Promise(() => iotMethodNotAllowed(result.$response.error))
+    }
+    if (result.$response.httpResponse.statusCode === 500) {
+      return new Promise(() => failure(result.$response.error))
+    }
+    return new Promise(() => ('Unknown territory'))
```
vdiezel: Nice! Now that I think of it we probably could even bypass the promise creation for the errors just like this:
```
if (result... === 400) {
  badRequest(...)
}
```
because those error functions throw an error, we do not need to return anything in that cases. For the happy path we  can just return the success string. We still can sill can ( and have to ) await the call in business logic like this :
`return await IotDataAPI.publishCommand(deviceId, payload)` which then will return only the success string in the successfull case. Or am I wrong here?

Also a small logical change: Instead of `if` statements use `if...else if` or a `switch` here.

vdiezel: And as the default case we could just use the `failure` instead of `Unknown territory` - I think sending a 500 to users in case something unpredictable happens is fine.

lib/response-lib.js
```diff
@@ -31,3 +31,13 @@ exports.failedValidation = (reason) => {
...
+
+exports.iotUnauthorized = (reason) => {
+  const message = `UnauthorizedException: ${reason}.`
+  throw new createError(401, message)
+}
+
+exports.iotMethodNotAllowed = (reason) => {
+  const message = `MethodNotAllowedException: ${reason}.`
+  throw new createError(405, message)
+}
```
vdiezel: new line!

lib/IotDataAPI.js
O-Susanoo: for the first maybe you are right but wouldn't it be weird if we await for a function that dose not return a promise  

for the if and else if , when there is a return statement it's not needed actually because the code will stop on the first success 

for the unknown territory part  i don't think 500 will do it because well it's not a 500 but maybe i can make another function that throw a custom error , in theory it should never reach this place but well

lib/IotDataAPI.js
```diff
@@ -1,12 +1,29 @@
 const AWS = require('aws-sdk')
+const { badRequest, iotUnauthorized, iotMethodNotAllowed, failure } = require('./response-lib')
 
 class IotDataAPI {
-
-  static execute(mqttMsg) {
+  static async execute(mqttMsg) {
```
DanielSeehausen: have a PR up soon that will update this rule in eslint!

O-Susanoo:
Ok thanks Dan for the peering

PR #143, opened 2019-11-24
added a new "or" validator and provided possibility for admin acces

### LINKED ISSUE #141: check 403 validations for the user and implement a pattern to inject admin access
child of #106

Our assert functions for authorization need to be checked after refactoring if the calls in the business logic actually are checking access to the right action. And we also wnt to provide proper test coverage here!

vdiezel:
- double checked 403 errors in the user service
- implemented a new validator function called `or` which takes multiple validators and passes validation when one of the validators passes. This way we can inject admin access 
- wrote two new validators `hasAdminAccess` and `hasDeveloperAccess`
- provided more tests to cover admin access / more checks for 404

closes #141


lib/validation-lib.js
```diff
@@ -47,6 +47,20 @@ exports.assertMany = (validators) => {
...
+    const results = []
+    validators.forEach(validator => {
+      results.push(validator())
+    })
+    for (const result of results) {
+      if (!result.invalid) return result
+    }
+    return results[0]
+  }
+}
```
vdiezel: New validator operator, which takes an array of valdators and return `{ invalid: false}` when one of the validator passes. Otherwise it will return the first failed validation result.

services/user/src/user/user.js
```diff
@@ -24,11 +24,20 @@ class User {
 
 
   async get({ requestingUserId, requestedUserId }) {
-    const requestedUser = await this.usersDB.findOne({ id: requestedUserId })
+    const users = await this.usersDB
+      .find({ id: { $in: [ requestingUserId, requestedUserId ] }})
+      .toArray()
+
+    const requestingUser = users.find(user => user.id === requestingUserId)
+    const requestedUser = users.find(user => user.id === requestedUserId)
```
vdiezel: We now also need to fetch the requestingUser to check if he/she has admin access

services/user/src/user/user.js
```diff
@@ -24,11 +24,20 @@ class User {
...
 
     assertMany([
       () => canFind(requestedUser, `User ${requestedUserId}`),
-      () => isMyResource(requestingUserId, requestedUserId)
+      () => canFind(requestingUserId, `User ${requestingUserId}`),
+      or([
+        () => isMyResource(requestingUserId, requestedUserId),
+        () => hasAdminAccess(requestingUser)
+      ])
```
vdiezel: `or` operator in use. Here we check if a user tries to fetch his own data OR if an admin tries to fetch information about somebody else. If one of those doesn't fail, it will be successfull.

services/user/test/user.test.js
```diff
@@ -259,8 +270,9 @@ describe('User Class', () => {
     test('should throw 403: Insufficient permission when user tries to delete a notification of somebody else', async() => {
       const notification = generateNotification()
       const user = await seedUser({ notifications: [ notification ] })
+      const requestingUser = await seedUser()
```
vdiezel: we now need a requestingUser also to exists in the databse, otherwise we will get a 404 instead of a 403

vdiezel:
In general I'd like to replace the whole role actions deal with either `hasAdminAccess` or `hasDeveloperAccess` in the other services too.

lib/validation-lib.js
```diff
@@ -47,6 +47,20 @@ exports.assertMany = (validators) => {
...
 }
 
+exports.or = (validators) => {
+  if (!Array.isArray(validators)) throw new Error('validators for \'or\' operator have to be an array')
+
+  return () => {
+    const results = []
+    validators.forEach(validator => {
+      results.push(validator())
+    })
```
DanielSeehausen: this is a good opportunity for a `map`
might also be able to make use of existing code :)
```js
const errorMsgs = validateMany(validators)
```

I haven't run this, but just eyeballing it, I am confused about the implementation. Will this not error [here](https://github.com/ospin-web-dev/hambda/blob/dc44ea60683ba7a4b3d62fa5bcbf9dbb304c0d4e/lib/validation-lib.js#L37)

when it is used [here](https://github.com/ospin-web-dev/hambda/pull/143/files#diff-7e80efd8982a2879fb5f6d57c1faed20R37)

- `assertMany` expects an array 
- it calls `assert(val)` on each value within that array
- `assert` throws error if its argument is not a function
- `or` (as it is now) does not return a function, but an object

##################################

related, but separate (I think we should use existing code, but this below is just putting some thoughts out there):
if we cleaned this up a step further, we could make it so `validateMany` doesn't return an array of:
```js
[null, null, 'errorMsg', null]
```
which was kinda sloppy on my part quite frankly
...but instead a compacted array, which would yield:
```js
['errorMsg']
```
this would make the next line of our function flow really naturally:

```js
const errorMsgs = validateMany(validators)
const allFailed = (errorMsgs.length === validators.length) 
...
```

services/user/src/user/user.js
```diff
@@ -24,11 +24,20 @@ class User {
 
 
   async get({ requestingUserId, requestedUserId }) {
-    const requestedUser = await this.usersDB.findOne({ id: requestedUserId })
+    const users = await this.usersDB
```
DanielSeehausen: [db has been abstracted](https://github.com/ospin-web-dev/hambda/blob/dc44ea60683ba7a4b3d62fa5bcbf9dbb304c0d4e/services/user/src/user/user.js#L15)

services/user/src/user/user.js
DanielSeehausen: this burns my heart, because its a lot more than

```js
const requestedUser = await this.usersDB.findOne({ id: requestedUserId })
const requestingUser = await this.usersDB.findOne({ id: requestingUserId })
```

but I am guessing you rightfully did it because our DB is not running on the same instance as our server, so ++ from me. Good thinking and way to take the extra step to do the right thing.

services/user/src/user/user.js
DanielSeehausen: see my comment above at the function declaration. I am unsure how this won't throw an error re: `or`'s return value

lib/validation-lib.js
vdiezel: The return value of or: `or` does not return an object, it returns a function that runs the validators. I considered `validateMany`  - but I would need to change the function to a function that does the exact same thing as our validators. The issue here is that `or` has to return a function that, when executed, returns an object like `{  invalid: .... }`, so I'm not just interested in the plain error message. Our validators do already exactly this. Putting a `validate` function in between is excessive imho. We are also not using them anywhere since you made that nice assertion pattern here, so I think I will just get rid of the `validate` and `validateMany` completely because they don't really serve any purpose anymore.

services/user/src/user/user.js
vdiezel: Yes, wanted to prevent going twice to the database.

services/user/src/user/user.js
vdiezel: `or` does return a function. ;)

DanielSeehausen: ima newb

lib/validation-lib.js
DanielSeehausen: thanks for explanation - totally missed that return line
this is an example why I shouldn't read a PR late at night!

PR #148, opened 2019-11-25
double checked 403 access in process service and added more tests

### LINKED ISSUE #147: double check 403 validations on the process service
child of #106

vdiezel:
closes #147 

- Modified process access validation to consider device access 
- added tests to check for device access level for modifed functions


services/process/src/process/process.js
```diff
@@ -101,11 +101,6 @@ class Process {
       assert(() => hasObjectAccess(process, requestingUserId, RESOURCE.PROCESS, PROCESS_ACTIONS.DELETE))
     }
 
-    // deleting datapoints
-    for (const processId of processIdsToDelete){
-      await this.dataPointsDB.deleteMany({ processId })
-    }
-
```
vdiezel: moved after all assertions so we don not modify any data before we didn't go through all assertions

services/process/src/process/process.js
```diff
@@ -21,7 +21,7 @@ class Process {
 
     assertMany([
       () => canFind(device, `Device ${deviceId}`),
-      () => hasObjectAccess(device, requestingUserId, RESOURCE.DEVICE, DEVICE_ACTIONS.GET)
+      () => hasObjectAccess(device, requestingUserId, RESOURCE.DEVICE, DEVICE_ACTIONS.OPERATE)
```
vdiezel: to be able to create, clone, update and delete processes, users now need to have `Execute` rights.

services/process/src/process/process.js
DanielSeehausen: 🕵

PR #153, opened 2019-11-26
removed hard-coded bucket reference and injecting the right env var

### LINKED ISSUE #152: Provide proper bucket references in requestStreamData and requestProcessData
currently, in the two functions `requestStreamData` and `requestProcessData` we are using a hard coded reference to the the `ospin-download-dev` bucket, which breaks in production because the assumed lambda role for those functions on prod has only access to `ospin-download-prod`. We need to inject the proper environmental variable `DOWNLOAD_BUCKET` here.

vdiezel:
Production env dowloads didn't work because the the bucket reference was always pointing to the S3 bucket of the dev envirnoment and the assumed lambda role doesn't have access to it.

closes #152 

**text for costumer to be added to some changelog PR**
Downloading process data or single stream data did not work as intended. For short processes and single streams this issue has been resolved. However, when process data exceeds a certain limit of data, downloads are cancelled by our backend provider. We are working on a solution.


services/process/serverless.yml
```diff
@@ -265,6 +265,7 @@ functions:
       DB_USER: ${{file(../../env.${{self:provider.stage}}.yml):DB_USER}}
       DB_PASSWORD: ${{file(../../env.${{self:provider.stage}}.yml):DB_PASSWORD}}
       DB_NAME: ${{file(../../env.${{self:provider.stage}}.yml):DB_NAME}}
+      DOWNLOAD_BUCKET: ${{file(../../env.${{self:provider.stage}}.yml):DOWNLOAD_BUCKET}}
```
vdiezel: the download bucket env var is stored in the `env` files (should be already there in your files)

PR #154, opened 2019-11-26
double check device authorization and add new validators

### LINKED ISSUE #149: double check 403 validations on the device service
child of #106

vdiezel:
- fixes a bug in updating stream defaults ( + test )
- double checked authorization calls
- implemented new `hasDeveloperAccess`authorizer instead of `hasSufficientRole`
- got rid of the `role` logic from the `access-rights-lib` - was an overkill

closes #149


DanielSeehausen:
image

this looks 🏅 
Given most of the + is from a new test, this is a nice reduction in code and complexity! love it. 
:shipit:

DanielSeehausen:
@O-Susanoo don't forget to turn on trim whitespace on save! see the test file in the diff for an example. (I know you have used something like 3 computers in the past two weeks, sorry to bother you to keep changing your vscode settings 😟 

might be a good time for me to push my evangelical mission of trying to convince people the value of version controlling their dotfiles! [vscode makes it easy!](https://code.visualstudio.com/docs/getstarted/settings#_settings-file-locations)

O-Susanoo:
Thanks for removing spaces from my tests , i will install an extension that show's me those spaces , if i can't use auto formatting

vdiezel:
np - vim removes my spaces automatically

PR #156, opened 2019-11-27
Eslint

DanielSeehausen:
error on my part!


DanielSeehausen:
> it would have been great if this would have been activated after a specific number of chars , i feel like with a function of two params it would be annoying at some point , but hee it's not like it will stop us from pushing

[I agree! I just looked for an option and I don't see one :(](https://eslint.org/docs/rules/comma-dangle)

PR #166, opened 2019-12-05
allowed empty string for a comment

### LINKED ISSUE #165: creating processes doesn't work when no comment for the process is provided
It fails validation because no empty comment is allowed

vdiezel:
```diff
! don't forget ticket number in branch!
```
closes #165 by allowing empty comment. Added a test to cover this.


services/process/test/process.test.js
```diff
@@ -160,6 +160,19 @@ describe('Process Class', () => {
       expect(fetchedDevice.processes[0]).toEqual(body.id)
     })
 
+    test('should create a process in the database when no comment provided', async() => {
+      const { user, device } = await seedUserDevice()
+      const process = generateProcess(user, device)
+
+      delete process.comment
```
O-Susanoo: deleting the comment means that the object will not have it i think that the test should be setting it to empty string ?

DanielSeehausen: good 👁 @O-Susanoo 
i'd second that. I wouldn't get rid of this test though - just add an additional for accepting empty string!

vdiezel: Yes, was a bit too quick there, you are right!

PR #167, opened 2019-12-06
367 change error message api

vdiezel:
Backend part of the changed error API.

The source of the error was that when the response from the server was a plain string and not a stringified JSON, due to the set `content-type: application/json` in the headers, axios always tried to call a JSON parse. And if this fails, it returns `null`. 

We could fix this by setting the right headers when errors occur to be `text/plain` and axios would provide the error string to the application. But I didn't like the idea of setting different content headers depending on the outcome of the request. 

We could therefore either use objects within the `createError` constructor 

`new createError(404, { message: { errorMessage: 'Device not found.' }})` 

(it has to be a nested object, because with `createError(404, { message })` the httpError-middleware would still assign the content of the message to the body, not the object `{ message: '...message' }`)

While we could do this, some prefab middleware uses plain strings in the constructor. 
So I modified our stringifyBody-middleware to react different in case of an error to create an object from the error message.

Second commit just fixes tests.


lib/middleware/stringifyResponseBody.js
```diff
@@ -8,11 +8,19 @@ const middleware = (handler, next) => {
...
+  })
+
+  return next()
 }
 
 
 module.exports = () => ({
   after: (handler, next) => middleware(handler, next),
-  onError: (handler, next) => middleware(handler, next)
+  onError: (handler, next) => errorMiddleWare(handler, next)
```
vdiezel: Only logic change of this PR. Rest is just fixing tests.

lib/middleware/stringifyResponseBody.js
```diff
@@ -8,11 +8,19 @@ const middleware = (handler, next) => {
   if (!bodyIsString(handler))
     handler.response.body = JSON.stringify(handler.response.body)
 
-  next()
+  return next()
+}
+
+const errorMiddleWare = (handler, next) => {
```
DanielSeehausen: middleware is one word, so camelcased: `errorMiddleware`

lib/middleware/stringifyResponseBody.js
DanielSeehausen: middleware is one word, so camelcased: `errorMiddleware`

DanielSeehausen: (not requesting a change)

vdiezel:
> so you did actually end up using the body , i guess it's better it's faster , however do you have in mind a client code that may use this code ? or that can be impacted by this change

Yep, here: https://github.com/ospin-web-dev/REACTor/pull/370

PR #170, opened 2019-12-11
Fix timeout issues : solution 1

### LINKED ISSUE #125: Spike: Investigate intermittent time out issues on backend functions
see lambdog warnings

O-Susanoo:
I found two possible solutions

a clean one but not sure about it's correctness , and this branch is for this solution , the main idea is that i suspect that there is a timeout issue or something that closes the db handler likely we have handlers that listens to this kind of events and this solution do exactly that it listens for events that can lead to a break in the connection and by doing so we re-instantiate the db handler

Closes #125


DanielSeehausen:
attempting modified version on `getUser`

O-Susanoo:
the docs for events are found here [Docs](http://mongodb.github.io/node-mongodb-native/3.1/api/Db.html#event)

O-Susanoo:
It's not working i guess beside the other solution provided before i don't see how we can overcome this issue , closing it for now

PR #172, opened 2019-12-17
changed updateProcess Joi schema to consider name validation

### LINKED ISSUE #168: don't allow empty process name on `update` process call

vdiezel:
closes #168 
Changes the `updateProcess` validation schema, so when a name is passed, it is being validated the same way as it is when we create a process


services/process/src/joi/process.js
```diff
@@ -51,7 +51,9 @@ const list = assertSchema(Joi.object({
 const update = assertSchema(Joi.object({
   requestingUserId: Joi.string().required().pattern(UUIDV4_REGEX),
   processId: Joi.string().required().pattern(UUIDV4_REGEX),
-  data: Joi.object()
+  data: Joi.object({
+    name: Joi.string().min(1).max(45),
+  }).unknown(),
```
vdiezel: This pattern allows to set any key in the data (the update object), but when name is passed, it is being validated.

O-Susanoo: Cool trick

DanielSeehausen: Are there any keys we don't want to allow to be updated?
Here is the create methods schema:

```js
{
      id: uuidv4(),
      deviceId: deviceId,
      deviceName: data.deviceName,
      createdAt: new Date().getTime(),
      createdBy: requestingUserId,
      name: data.name,
      cloneHistory: data.cloneHistory || [],
      description: data.description,
      entryPhaseId: data.entryPhaseId,
      progression: [],
      comment: data.comment,
      ownerId: requestingUserId,
      deviceDescription: data.deviceDescription,
      access:  createDefaultGroups('process', requestingUserId),
      startedBy: null,
      startedAt: null
}
```

vdiezel: Yes, some of them should be able to be updated. I'll check out Joi what I can do here.

vdiezel:
changed my approach on the schema and made it exclusive, now you got to define the keys in the schema you want to allow to be modified.

PR #174, opened 2019-12-18
398 store process finish time

### LINKED ISSUE #173: Modify backend to store the finish time of a process

vdiezel:
closes #173 

Now stores a `finishedAt` property on the process when a device goes back to idle after running a process. Main change in device/iotTrigger/updateStateTrigger.

Also implemented tests for this trigger and by doing this I added a new top level testing helper called `Seeder` that we can now reuse from all services.


services/device/src/iotTrigger/trigger.js
```diff
@@ -73,11 +73,27 @@ class Trigger {
...
+    // update device state
+
     await devices.updateOne({ id: deviceId }, {
       $set: {
-        state: state,
-        runningProcessId: processId || null
-      }
+        state,
+        runningProcessId: processId || null,
+      },
```
vdiezel: The only logic change to the source code.

test/helper/Seeder.js
```diff
@@ -0,0 +1,66 @@
...
+    user.processes.push(processId)
+    user.devices.push(device.id)
+
+    await this.usersDB.insertOne(user)
+
+    return { user, device, process }
+  }
+}
+
+module.exports = Seeder
```
vdiezel: This class now takes care of seeding the database for our test runner.

services/device/src/iotTrigger/trigger.js
O-Susanoo: Nice job,

i have one remark tough (listening to the inner voice of SOLID way :) ) it seem to me that this function is doing a little bit more than what it promise to do maybe in the future add a function called finishProcess or something like this, or in case of being lazy rename the updateState to updateStateAndSetFinishProcess , but nothing blocking

services/device/src/iotTrigger/updateState.js
```diff
@@ -35,8 +35,5 @@ async function updateState(event) {
 
 exports.main = middy(updateState)
   .use(logUnhandledErrors())
-  .use(httpErrorHandler({ logger }))
   .use(doNotWaitForEmptyEventLoop())
-  .use(stringifyResponseBody())
-  .use(cors())
```
vdiezel: removed those because this Lambda function is triggered by the rules engine triggered via MQTT

services/device/src/iotTrigger/updateState.js
```diff
@@ -35,8 +35,5 @@ async function updateState(event) {
```
O-Susanoo: as talked about , great job removing unused middlewares but i see they are still being imported

services/device/test/trigger/updateState.test.js
```diff
@@ -0,0 +1,107 @@
...
+        )
+        await fireUpdateStateEvent(device.id, processId, newState)
+        const updatedDevice = await dbSeeder.devicesDB.findOne({ id: device.id })
+
+        expect(updatedDevice.state).toBe(newState)
+        expect(updatedDevice.runningProcessId).toBe(processId)
+      }
+    })
+  })
+})
```
O-Susanoo: what will happen if someone clone a process will it also close the finishedAt ?

test/helper/Seeder.js
O-Susanoo: Cool

services/device/src/iotTrigger/updateState.js
O-Susanoo: as discussed , great job removing unused middlewares but i see they are still being imported

services/device/test/trigger/updateState.test.js
vdiezel: Great catch! Yes, we need to clear this key. We should probably refactor cloning to take only specific keys, instead of overwriting abundant keys from the copy of the original process here. Lesson learned.

lib/middleware/logUnhandledErrors.js
```diff
@@ -10,7 +10,7 @@ module.exports = (opts) => {
     onError: (handler, next) => {
       const isHttpError = handler.error.statusCode
 
-      if (!isHttpError && typeof options.logger === 'function') {
+      if (!isHttpError && typeof options.logger === 'function' && process.env.NODE_ENV !== 'test') {
```
vdiezel: silencing the logger in test mode

lib/validation-lib.js
```diff
@@ -16,20 +16,28 @@ const { group, ROLES } = require('access-rights-lib')
...
+  }
 }
 exports.assert = assert
 
-exports.assertMany = (validators) => {
+exports.assertMany = (validators, { noHTTP = false } = {}) => {
   if (!Array.isArray(validators)) throw new Error('validators have to be an array')
 
-  validators.forEach(assert)
+  validators.forEach(validator => assert(validator, { noHTTP }))
```
vdiezel: considers now a `noHTTP` option to overwrite the response function with a normal Error so we can reuse our pattern in lambda handlers which are not triggered by HTTP calls.

services/device/src/iotTrigger/trigger.js
```diff
@@ -70,13 +70,32 @@ class Trigger {
...
+        finishedAt: Date.now(),
+      },
+    })
+  }
+
+  async handleStateUpdates({ processId, deviceId, state }) {
     const devices = this.db.collection('devices')
+    const device = await devices.findOne({ id: deviceId })
+
+    assert(() => canFind(device, `Device ${deviceId}`), { noHTTP: true })
```
vdiezel: usage of the noHTTP option

services/device/src/iotTrigger/updateState.js
```diff
@@ -22,21 +16,17 @@ async function updateState(event) {
...
 
-  const result = await trigger.updateState(requestData)
+  const result = await trigger.handleStateUpdates(requestData)
   return success(result)
 }
 
 exports.main = middy(updateState)
-  .use(logUnhandledErrors())
-  .use(httpErrorHandler({ logger }))
+  .use(logUnhandledErrors({ logger }))
```
vdiezel: added the env dependent logger to this so its silenced when errors are thrown in `test`

PR #179, opened 2019-12-29
134 transfer device ownership

### LINKED ISSUE #178: Create a transfer ownership endpoint for device
child of #134 
Device backend part

vdiezel:
closes #178 

Provides the possibility to transfer the ownership of a device to another device user.
Comments provided where important. Most file changes due to renaming of a couple of functions in validation-lib.


lib/validation-lib.js
```diff
@@ -85,7 +85,20 @@ exports.hasObjectAccess = (resource, userId, resourceType, action) => {
...
+
+  if (!resourceUserGroup) {
+    return {
+      invalid: true,
+      errorMsg: `User ${userId} has no access to resource ${resource.id}`,
+      responseFn: noAuthorization
+    }
+  }
+  return { invalid: false }
+}
```
vdiezel: New validator function to see if a user is present on a resource.

lib/validation-lib.js
```diff
@@ -64,7 +64,7 @@ exports.canFind = (resource, label = 'Resource') => {
   return { invalid: false }
 }
 
-exports.hasObjectAccess = (resource, userId, resourceType, action) => {
+exports.hasSufficientResourceAccess = (resource, userId, resourceType, action) => {
```
vdiezel: renamed this one to be more on point

lib/validation-lib.js
```diff
@@ -96,6 +109,17 @@ exports.isMyResource = (requestingUserId, targetUserId) => {
...
+  if (requestingUserId !== resource.ownerId) {
+    return {
+      invalid: true,
+      errorMsg: `User ${requestingUserId} is not the owner of resource ${resource.id}`,
+      responseFn: noAuthorization
+    }
+  }
+  return { invalid: false }
+}
+
```
vdiezel: new validator function that checks if a user owns a resource.

services/device/src/device/device.js
```diff
@@ -277,13 +277,50 @@ class Device {
...
+      bulk.find({ id: deviceId, 'access.name': 'owners' }).updateOne( {
+        $push: { 'access.$.users':  newOwnerId }
+      })
+
+      bulk.find({ id: deviceId, 'access.name': newOwnersOldGroupName }).updateOne( {
+        $pull: { 'access.$.users':  newOwnerId }
+      })
+
+      await bulk.execute()
+    }
```
vdiezel: the new user will be put into the owners group if he Isn't already in there.

vdiezel: The general idea is this: I can tranfer ownership if I'm the owner of the device and if the new owner has already device access.

lib/validation-lib.js
```diff
@@ -85,7 +85,20 @@ exports.hasObjectAccess = (resource, userId, resourceType, action) => {
...
+  const resourceUserGroup = resource.access.find(group => group.users.includes(userId))
+
+  if (!resourceUserGroup) {
+    return {
+      invalid: true,
+      errorMsg: `User ${userId} has no access to resource ${resource.id}`,
+      responseFn: noAuthorization
+    }
+  }
+  return { invalid: false }
```
DanielSeehausen: invert order here to avoid `!`

lib/validation-lib.js
```diff
@@ -85,7 +85,20 @@ exports.hasObjectAccess = (resource, userId, resourceType, action) => {
...
+}
+
+exports.isMe = (requestingUserId, targetUserId) => {
```
DanielSeehausen: 🕵️‍♀️

services/device/src/device/device.js
```diff
@@ -277,13 +277,50 @@ class Device {
...
+    ])
+
+    assertMany([
+      () => canFind(device, `Device ${deviceId}`),
+      () => canFind(newOwner, `User ${newOwnerId}`),
+      () => isMyResource(requestingUserId, device),
+      () => hasResourceAccess(device, newOwnerId)
+    ])
+
+    const newOwnersOldGroupName = device.access.find(group => group.users.includes(newOwnerId)).name
```
DanielSeehausen: we are doing this in a couple places now. 
Next time one of works on it we should consider adding a method along the lines of `device.getUserGroup(device, user)` to standardize the way these things are accessed

DanielSeehausen: (or in `access-rights-lib` or somewhere else)

services/device/src/device/device.js
```diff
@@ -277,13 +277,50 @@ class Device {
...
+      })
+
+      await bulk.execute()
+    }
+
+    const result = await this.devicesDB.updateOne({ id: deviceId }, {
+      $set: { ownerId: newOwnerId }
+    })
+
+    return result.modifiedCount
```
DanielSeehausen: given that we have multiple modification actions happening here, might make sense to return a richer response that confirms success on both of the actions. Could be a string along the lines of:
`User: <id> is now the owner of device: <id>`
(with the potentially extra line of:)
`User: <id> is now the owner of device: <id> and has been added to the <owner/admin> group`

services/device/test/device/transferOwnership.test.js
```diff
@@ -0,0 +1,163 @@
...
+          const newOwner = await dbSeeder.seedUser()
+          await dbSeeder.addNewUserToDevice(newOwner.id, device.id)
+          const { res, body } = await fireTransferOwnershipEvent({ id: badId }, user, newOwner)
+          expect(res.statusCode).toBe(422)
+          assertContainsAll(body.message, [ 'deviceId' ])
+        }
+      })
+
+      test('should throw a 422: When newOwnerId does not match the Joi schema', async() => {
+        const badIds = [13245, undefined, 'so it goes']
```
DanielSeehausen: g4m3r

services/device/src/device/device.js
vdiezel: I agree. But because we have this going on in so many places, we need a standardized way to handle this. I created a new issue for this: https://github.com/ospin-web-dev/hambda/issues/190

vdiezel:
> looks awesome. tests were A+. I should have read them first then the function.
> 
> ! Only request is to get eslint running on this code and lint the new stuff added before merging in

I'll do this in a seperate branch! Seems ALE is not running for some reasons on the hambda repo. I will also modify the slint.rc so it matches the one from REACTor as much as possible.

PR #188, opened 2020-01-08
add -500 to null script converter

### LINKED ISSUE #185: Update `-500` sensor data to `null`
**This is to be completed AFTER #184 has been merged to both dev and prod**

Currently, the database has a lot of -500 values to represent bad data values for process datapoints 

write a utility that can be run on our mongo servers to crawl ALL datapoints and convert -500s to null values. 

While we only have one customer in production, this is still the most serious part of this ticket, being a massive update to existing data. Take a look into running scripts on mongo atlas/the mongo console for ideas on how to execute whatever script is made. The script should have error handling and recovery, and please write tests for it that run on some sample process data you copy from the DB before running it for real.

Be sure to run on dev on a small section first and assert its working in all cases. 
Please copy all of Marcus' processes on prod (FU/Berlin reactor or whatever it is called) over to dev and then run this utility on them and make sure they are loading no problem in dev.

O-Susanoo:
Add scripts folder to hambda 
add -500 to null script converter

closes #185


scripts/test/seeds/dataPoints.js
```diff
@@ -0,0 +1,53 @@
+exports.dataPoints = [
+  {
+    _id: '5d958f6750895389ec46e678',
```
DanielSeehausen: This looks like a good start:

The seed should be generic - take a look at how `generateProcess` is done (it can accept data which overrides its default data at the `...data` line. (ignore the `applyTraits` stuff - no need to copy that over at the moment)

Consider making this also `generateDataPoint` in order to match how we already do this across our code base (then in the tests, you generate seeds with optional data, and put them into mongo. See how other tests do it for reference). This has the added benefit that you see the optional data you are then asserting on in the test.

`_id`, being an internally assigned mongo value, shouldn't be present on the seed
`processId` should also not be hard coded here. I am unsure what `sid` is, but perhaps that also doesn't belong depending on what it is referencing?

Other than that it looks very good. If you are thinking in another direction or if this reasoning doesn't make sense please let me know. I think you may find test comprehension will go up once you can describe data values when you seed, as it keeps you from referencing defaults written in another file!

scripts/test/convertMinus500DataPointsToNull.test.js
```diff
@@ -0,0 +1,35 @@
...
+  })
+
+  test('should change all occurrences of -500 to null', async () => {
+    const idsBefore = await convert500toNull.getDataPointsWithNullValuesIds(db)
+    expect(idsBefore.length).toBe(2)
+    await convert500toNull.migrateDb(db)
+    const idsAfter = await convert500toNull.getDataPointsWithNullValuesIds(db)
+    expect(idsAfter.length).toBe(0)
+  })
+})
```
DanielSeehausen: these tests look like a good start!

if I could only write one test, this would also be mine. That being said, asserting the amount of occurrences in the database is too broad to feel comfortable with. 

I actually started by reading your seed value, and after seeing the defaults:
```
values: [-500, -500, -500]
values: [10.97, 11.97, 12.97]
values: [13.97, -500, 14.97]
```
It seemed you were setting up for additional test cases!
What jumps out immediately is:
 - to test that editing mixed dataPoints only edits those that are -500
 - test that arrays are the same length
 - test that order does not change
(All of these could be covered by asserting each dataPoint group has the right result in the end, rough e.g. 
```js
const dataPoint = seedDataPoint([
  {
    times: [
      29685120,
      29686097,
      29687132,
    ],
    values: [
      13.97,
      -500,
      14.97,
    ]
  }
], ... )

expect(dataPoint[0].values).toEqual([13.97, null, 14.97])
```

TL;DR: there is room for more coverage in these tests! See comment on the seed for how to make it very easy and user friendly to do that

scripts/test/convertMinus500DataPointsToNull.test.js
```diff
@@ -0,0 +1,35 @@
...
+    db = await mongo.db()
+    dataPointsDB = await db.collection('dataPoints')
+  })
+
+  afterAll(async () => {
+    await mongo.conn().close()
+  })
+
+  beforeEach(async ()=>{
+    await dataPointsDB.insertMany(dataPoints)
```
DanielSeehausen: see comment on seed file. 
should you expand these tests, this pattern locks us into asserting on default data on the seed that is obscured in another file. 
If the seed instead becomes a `generateSeed`, we can explicitly pass in the dataPoint values we would like to test against in the tests, which becomes really convenient

O-Susanoo:
Well , i actually didn't go throw all of that for 2 reasons , one this stuff is supposed to be a dev tool not a tool for backend code , but i can understand what you are saying , the second reason , after committing the code i realized that we can do the hole thing with just one mongo command

``` 
db.dataPoints.updateMany({ values: -500},{$set: { "values.$[elem]": null }},{ arrayFilters: [{ "elem": -500 }],multi: true })
```
So i don't know if we want to go through each record one by one anymore ?

scripts/convertMinus500DataPointsToNull.js
```diff
@@ -0,0 +1,36 @@
+async function updateDataPoints(db, dataPointId) {
+  await db.collection('dataPoints').updateOne({
+    _id: dataPointId._id,
+  }, {
+    $set: { "values.$[elem]": null }
+  }, {
+    arrayFilters: [{ "elem": -500 }],
+    multi: true
+  })
+  console.log(` Updating datapoint id : ${dataPointId._id}`)
```
DanielSeehausen: this logs in the test file :(, which we want to run cleanly.

image


You can fix that by overwriting `console.log` for only that test, or some other method. That is just what comes to mind.

scripts/convertMinus500DataPointsToNull.js
```diff
@@ -0,0 +1,36 @@
...
+  return db.collection('dataPoints').find().project({ _id: true }).toArray()
+}
+
+async function migrateDb(db) {
+  const dataPointsDbIds = await getDataPointsWithNullValuesIds(db)
+  for (let dataPointId of dataPointsDbIds) {
+    try {
+      await updateDataPoints(db, dataPointId)
+    } catch (error) {
+      console.error(` Failed to update datapoint id : ${dataPointId._id}`)
```
DanielSeehausen: we may want to consider writing this to a test file instead. That way, we have something to progromatically traverse in the event of error!

this is a great opportunity to make use of your computer's `/tmp` directory!

scripts/convertMinus500DataPointsToNull.js
```diff
@@ -0,0 +1,36 @@
...
+    }
+  }
+}
+
+module.exports = {
+  migrateDb,
+  getAllDataPoints,
+  getDataPointsWithNullValuesIds
+}
```
DanielSeehausen: We need to be able to run this on sections of the database (e.g. a single process, a whole device, all of an environment, etc.) One way to do this is to make sure the script can accept these things are argument (i.e. a process id, or device id). 

We discussed in person, but [I could have been more specific in the ticket](https://github.com/ospin-web-dev/hambda/issues/185), which does touch on this

scripts/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "scripts": {
+    "seed-lib-files": "cp -rf ../lib/* ./node_modules/",
+    "pretest": "npm run seed-lib-files",
+    "test": "jest -i"
```
DanielSeehausen: do we need the `-i` flag?
very well might, I was just unsure looking at it like this

O-Susanoo:
> Well , i actually didn't go throw all of that for 2 reasons , one this stuff is supposed to be a dev tool not a tool for backend code , but i can understand what you are saying , the second reason , after committing the code i realized that we can do the hole thing with just one mongo command
> 
> ```
> db.dataPoints.updateMany({ values: -500},{$set: { "values.$[elem]": null }},{ arrayFilters: [{ "elem": -500 }],multi: true })
> ```
> 
> So i don't know if we want to go through each record one by one anymore ? what do you think is it worth it to still preserve such script ?

and i thinks this option is the safest one , because we outsource the load on us and give it to mongodb and it's up to him to manage the hole thing

scripts/convertMinus500DataPointsToNull.js
vdiezel: The documents in the dataPoints collection are actually data chunks (yes, it's inconsistent with the name of the collection). So you might want to adjust the message from 'updating datapoint' to 'updating data chunk'

scripts/convertMinus500DataPointsToNull.js
```diff
@@ -0,0 +1,36 @@
...
+    multi: true
+  })
+  console.log(` Updating datapoint id : ${dataPointId._id}`)
+}
+
+async function getDataPointsWithNullValuesIds(db) {
+  return db.collection('dataPoints').find({ values: -500 }).project({ _id: true }).toArray()
+}
+
+async function getAllDataPoints(db) {
```
vdiezel: does this one returns also only ids? If so, this should be reflected in the naming!

scripts/test/seeds/dataPoints.js
vdiezel: `sid` stands for streamId. It's an artifact from when we wanted to store each message individually and wanted to save a couple of bytes.

PR #189, opened 2020-01-09
Transform -500 values to null on datapoint insert

### LINKED ISSUE #184: Convert bad/no data (500s) to null values
Currently, -500s are emitted from the device for bad/no/error values from sensors

update the back end to convert -500 into null values automatically

This will (hopefully) no longer be needed later when someone gets the time to do the backend work to send a proper value. Until then, we need some utility that is abstracted (ideally as an exported function) that parses the incoming data from the device and does this.

Please provide test coverage!

O-Susanoo:
I should mention that after this dev, now when user request to download datapoints they will receive a csv file with null in it , i think what we can do is omit those values when creating the csv file , but i didn't do that, i think we should confirm that with Jan

closes #184


lib/nullifyInvalidData.js
```diff
@@ -0,0 +1,15 @@
...
+  }
+*/
+module.exports = (streamsData) => (
+  Object.assign({}, ...Object.entries(streamsData).map(([ streamId, [ times, values ] ]) => {
+    const nullifiedValues = values.map((value) => {
+      if (value === -500) return null
+      return value
+    })
+    return { [streamId]: [times, nullifiedValues] }
+  }))
```
vdiezel: We can simply use destructruing here instead of `Object.assign`

DanielSeehausen:
> I should mention that after this dev, now when user request to download datapoints they will receive a csv file with null in it , i think what we can do is omit those values when creating the csv file , but i didn't do that, i think we should confirm that with Jan
> 
> closes #184

Good catch @O-Susanoo, and I think you are 100% right to recognize this is the type of thing we shouldn't be deciding as the engineers. This was actually already brought up (I discussed last friday with Jan) and for now it is ok/maybe right that we include the null values (they are certainly way better than -500). The thought is, we want user's downloadable data to reflect the real data as much as possible for now, and the `null` does actually communicate some potentially important information (being, the sensor reported, it just didn't report any real data). Things like excel can gracefully handle those values when charting data!

Perhaps it is something we remove down the line though. I think it is likely we will want more specific null values from firmware though, (i.e. was it an error/bad sensor, vs no data, vs sensor was unplugged, etc.)

DanielSeehausen:
> I like the solution! Regarding the null values in the downloaded csv files: It's ok for now. Currently they receive -500s instead, which is just as bad so we do not make the situation worse. We plan to have a more sophisticated solution for donwloading data and there we can provide an option that allows the user to set a flag so null values are filtered out.

just saw this now...

lib/nullifyInvalidData.js
```diff
@@ -0,0 +1,15 @@
...
+)
```
DanielSeehausen: `lib` is generally for shared stuff across services
I believe this will only be used in `device` (and maybe `trigger` if/when we one day split that out as a service)

potential place for it is in `device/src/serializers`!

services/device/src/iotTrigger/trigger.js
```diff
@@ -1,6 +1,6 @@
 const IotDataAPI = require('IotDataAPI')
 const { assert, canFind } = require('validation-lib')
-
+const nullifyInvalidData = require('nullifyInvalidData')
 const serializeStreamsData = require('../serializers/serializeStreamsData')
```
DanielSeehausen: the space that was here was breaking absolute and relative imports
no biggie, just wanted to point out the reason for it before in case this is the first time you are seeing it

lib/nullifyInvalidData.js
DanielSeehausen: in fact, could just add the function into `serializeStreamsData` and then call it in the main function in there

lib/nullifyInvalidData.js
O-Susanoo: I don't know how you can do that ? any ideas

lib/nullifyInvalidData.js
O-Susanoo: i thought about that also but i didn't want to pollute the serializes folder , plus the sort function is also there , so i guess it's appropriate for it to be also there ?

services/device/test/trigger/event-generators/createDataPointsToDbEvent.js
```diff
@@ -0,0 +1,7 @@
+const createDataPointsToDbEvent = (messageSequence, processId, streamsData) => ({
+  messageSequence,
+  processId,
+  ...streamsData
```
DanielSeehausen: is the spread because we don't want someone messing with the original array without realizing they are changing the generated event's internal data as well?

services/device/src/iotTrigger/trigger.js
O-Susanoo: nice catch

services/device/test/trigger/event-generators/createDataPointsToDbEvent.js
O-Susanoo: no it's there because data is apparently suppose to arrive like this , i sent quite a while yesterday to understand that i needed to add this to the Event creator , because if i just put streamsData then i will have a streamsData object and this is not how the function expect data to be , because it's just a device that's sending this stuff it's not too pretty

services/device/test/trigger/dataPointsToDb.test.js
```diff
@@ -0,0 +1,80 @@
...
+
+    await fireDataPointsToDbEvent(1, processId, streamData)
+
+    const savedRecords = await db.collection('dataPoints').find().toArray()
+
+    savedRecords.forEach(row => {
+      assertTransformation(row, streamData[row.sid])
+    })
+  })
+})
```
DanielSeehausen: test file looks awesome!

DanielSeehausen: (we pair on one small QoL optimization once back from lunch)

PR #199, opened 2020-01-21
198 allow device admins to crud all processes

### LINKED ISSUE #198: Allow admins of devices to retrieve and modify all processes

O-Susanoo:
closes #198


lib/validation-lib.js
```diff
@@ -193,3 +196,27 @@ exports.jsonMatchesSchema = jsonMatchesSchema
 exports.assertSchema = (schema) => (
   (data) => assert(() => jsonMatchesSchema(schema, data))
 )
+
+const isDeviceAdmin = (device, userId) => {
+  const resourceUserGroup = device.access.find(group => group.name === 'owners' && group.users.includes(userId))
```
DanielSeehausen: couple thoughts here - would love to know what you think

this might be better suited as a `isResourceAdmin` to stick with the duck typing pattern @vdiezel created. 

we have been on a steady path of abstracting these iterator look ups into single functions to standardize our API into these models. This one looks like a pretty usual suspect for 'there should only be one way (function) to ask if someone is an admin on a device/resource. 

To build on that, the desire for this function to be abstracted really jumps out if you ask the question: "what is the role of the validation-lib?" It might be fair to say device access implementation should not be owned here. Unfortunately, the best other contender (access-rights-lib) isn't yet recreating this kind of behavior. I am wondering if we are taking on technical debt by continuing to expand `validation-lib` with things that are outside of its job (managing validations/validators), and if there is a cheap way to get ahead of it now, instead of going back and yanking stuff out later.

lib/validation-lib.js
```diff
@@ -193,3 +196,27 @@ exports.jsonMatchesSchema = jsonMatchesSchema
...
+  return {
+    invalid: true,
+    errorMsg: 'User has no admin access over device',
+    responseFn: noAuthorization
+  }
+}
+
+exports.isDeviceAdmin = isDeviceAdmin
+
+exports.hasProcessAccess = (device, process, userId, action) => (
```
DanielSeehausen: considering the arguments passed in, and that it is ambiguous with a device/process/user all passed in, consider `userHasProcessAccess`

services/device/src/device/device.js
```diff
@@ -492,6 +502,24 @@ class Device {
...
+  async getProcesses({requestingUserId, deviceId }) {
+    const device = await this.devicesDB.findOne({ id: deviceId })
+    const user = await this.usersDB.findOne({ id: requestingUserId });
+
+    assertMany([
+      () => canFind(device, `Device ${deviceId}`),
+      () => canFind(user, `User ${requestingUserId}`),
+    ])
+
+    const processIds = isDeviceAdmin(device, requestingUserId).invalid ? user.processes : device.processes
```
DanielSeehausen: assuming a users `processes` array is an accurate array of process ids, consider what processes this function is going to return if a user is not a device admin. We may want to be more accurate here.

services/device/test/device/seeds/device.js
```diff
@@ -61,11 +61,10 @@ function applyTraits(originalData, traits, user) {
 }
 
 const generateDevice = (user, data = {}, traits = []) => {
-
   const deviceData = {
     id: uuidv4(),
     name: `${faker.company.companyName()} Device`,
-    access: createDefaultGroups('device', user.id),
+    access: createDefaultGroups('device', data.deviceData && data.deviceData.userIsNotOwner ? undefined : user.id),
```
DanielSeehausen: if a line is getting crazy like this (key assignment is a function has a ternary in its second arg), it can help future readers by breaking the logic out.

DanielSeehausen: would the 'USER_AS_OWNER' trait up above in this file help you here? On first glance, we might be updating the seed generator to match a tests implementation

lib/validation-lib.js
```diff
@@ -193,3 +196,27 @@ exports.jsonMatchesSchema = jsonMatchesSchema
...
+  or([
+    () => hasSufficientResourceAccess(process, userId, RESOURCE.PROCESS, action),
+    () => isDeviceAdmin(device, userId),
+  ])
+)
```
DanielSeehausen: this guy breaks our api and user expectations - now validators return un-invoked functions!
should this return the result of the or, or the or itself?

services/process/src/dataPoints/dataPoints.js
```diff
@@ -9,15 +9,20 @@ const util = require('util')
...
 
+    assert(() => canFind(process, `Process ${processId}`))
+
+    const device = await this.devicesDB.findOne({ id: process.deviceId })
+
     assertMany([
-      () => canFind(process, 'Process'),
-      () => hasSufficientResourceAccess(process, userId, RESOURCE.PROCESS, PROCESS_ACTIONS.GET_PROCESS_DATA)
+      () => canFind(device, 'Device'),
+      () => hasProcessAccess(device, process, userId, PROCESS_ACTIONS.GET_PROCESS_DATA)(),
```
DanielSeehausen: left a comment on this being invoked here

services/process/src/process/process.js
```diff
@@ -56,16 +56,14 @@ class Process {
...
+    assert(() => canFind(process, `Process ${processId}`))
 
-    const device = await this.devicesDB.findOne({ id: process.deviceId });
+    const device = await this.devicesDB.findOne({ id: process.deviceId })
 
     assertMany([
       () => canFind(device, `Device ${process.deviceId}`),
-      () => hasSufficientResourceAccess(device, requestingUserId, RESOURCE.DEVICE, DEVICE_ACTIONS.GET)
+      () => hasProcessAccess(device, process, requestingUserId, PROCESS_ACTIONS.GET)(),
+      () => hasSufficientResourceAccess(device, requestingUserId, RESOURCE.DEVICE, DEVICE_ACTIONS.GET),
```
DanielSeehausen: I believe a user making a request to the GET `process/<id>` endpoint is supposed to get back their resource, if they are authorized for that resource. In this case, we are now rejecting if they aren't authorized for the device of the process. If I am missing a bigger picture here, please let me know. 

If there isn't a test case covering this, one should be added!

services/process/src/process/process.js
```diff
@@ -99,28 +97,26 @@ class Process {
     const foundProcesses = await this.processesDB
       .find({ id: { $in: processIdsToDelete }})
       .toArray()
+    const targetDeviceIds = [
+      ...new Set(foundProcesses.map(foundProcess => foundProcess.deviceId))
+    ]
+
+    const targetDevices = await this.devicesDB.find({ id: { $in: targetDeviceIds }}).toArray()
```
DanielSeehausen: understanding we are now running a call to devicesDB in all of our process validators for the full device, is there any way we could combine these calls? Just spitballing options here:

1. since this is done often now, we can consider adding a model method (or model db helper method). Something like `getProcessWithDevice` (`getProcessesWithDevices`), which would do a lot to help keep the model functions that effect change/return data lean. 
2. mongoose has `.populate()`. Is there anything like that we can use to autofill a `device` onto the `process` object when the fetch occurs?
3. is there any way to make a compound call, where it examines the process deviceId and then returns an additional device document it loads?

services/device/src/device/device.js
```diff
@@ -492,6 +502,24 @@ class Device {
 
     return result.modifiedCount
   }
+
+  async getProcesses({requestingUserId, deviceId }) {
```
vdiezel: missing space after curly brace :-)

services/device/src/device/device.js
vdiezel: I agree. `user.processes` contains references to all processes a user owns, independent of the device.

services/device/test/device/seeds/device.js
vdiezel: `Seeder` allows passing in the traits. Haven't tested it yet, but we should make use of this nice pattern!

services/process/src/process/process.js
vdiezel: The idea here is that a user may only get access to process on a device when they also have at least `GET` device access. It is to cover this case: A user has several processes they have created on the device. A device admin decides to remove the user from the device. Should we still allow access to the processes for the removed user? or now, we decided: No.

We probably need a bigger discussion about the authorization here.

services/process/src/process/process.js
vdiezel: I'm in favour of looking for a mongo function that does something like `populate` (maybe `$lookup` ?).

services/process/test/process/delete.test.js
```diff
@@ -157,7 +157,7 @@ describe('deleteProcess', () => {
 
     describe('when the requesting user is in the \'owners\' access group of some processes requested for deletion', () => {
       test('should throw 403 and a meaningful message, while deleting none of the processes requested', async() => {
-        const { user, device, process } = await dbSeeder.seedUserDeviceProcess()
+        const { user, device, process } = await dbSeeder.seedUserDeviceProcess(undefined, { deviceData: { userIsNotOwner: true } })
```
DanielSeehausen: see comment in `services/device/test/device/seeds/device.js, as this may be duplicate work to functionality already provided (using traits)

services/process/test/process/process.test.js
```diff
@@ -84,9 +84,9 @@ describe('Process Class', () => {
     return { user, device }
   }
 
-  async function seedUserDeviceProcess(processData = {}, traits = []) {
+  async function seedUserDeviceProcess(processData = {}, traits = [], deviceData = {}) {
```
DanielSeehausen: This might be able to go now. Using the db seeder replaces this. same as you [are using it now in device test](https://github.com/ospin-web-dev/hambda/pull/199/files#diff-0219ee334b4c8ed1036513e83aefbac0R160)

DanielSeehausen: (the args might be out of whack in terms of order/arity compared to dbseeder and the former duplicate implementation in device.test, and that is my fault!)

test/helper/generators/device.js
```diff
@@ -61,11 +61,10 @@ function applyTraits(originalData, traits, user) {
 }
 
 const generateDevice = (user, data = {}, traits = []) => {
-
   const deviceData = {
     id: uuidv4(),
     name: `${faker.company.companyName()} Device`,
-    access: createDefaultGroups('device', user.id),
+    access: createDefaultGroups('device', data.userIsNotOwner ? undefined : user.id),
```
DanielSeehausen: `services/device/test/device/seeds/device.js` re: this usage.

looking at this file, I believe it is a duplicate (or very close to a duplicate) of services/device/test/device/seeds/device.js

this is the top level one we should be using, and the other should be 💥 
any functionality not present in this one that is in services/device/test/device/seeds/device.js should likely be pulled into this one so the tests don't need to change!

services/process/src/process/process.js
DanielSeehausen: this explanation 💯% makes sense to me under our current given constraints. Sounds great.

A bigger discussion sounds good one day (when a user removes another from a device, do they also check a box which says remove them from processes, or is it done by default, etc.), but who knows when that day will be needed

services/device/src/device/device.js
DanielSeehausen: :laser: 👀

lib/validation-lib.js
O-Susanoo: i remember you suggesting the name hasProcessAccess but well i will changed

services/device/src/device/device.js
O-Susanoo: thanks for the note i completely overlook it

services/process/src/process/process.js
O-Susanoo: i'm ok with a helper function that returns both process and device , but i think it's a bad idea to merge device or process with each other , it seem messy cause we are no longer dealing with objects as we get theme from database

lib/validation-lib.js
```diff
@@ -193,3 +196,27 @@ exports.jsonMatchesSchema = jsonMatchesSchema
 exports.assertSchema = (schema) => (
   (data) => assert(() => jsonMatchesSchema(schema, data))
 )
+
+const isResourceAdmin = (resource, userId) => {
+  const resourceUserGroup = resource.access.find(group => group.name === 'owners' && group.users.includes(userId))
```
DanielSeehausen: this is looking a lot better!

minor room for clarification:
consider a developer later reading this function line by line --
when they get to this line, they see we are finding a resource group and assigning it out. They should then make the assumption "this function will later make use of this resource access group we found".
what we actually use it for is to ask if the user was found in the resource group 'owners', and don't end up doing anything with the group

an easy fix is to change the iterator to `some`, which returns a bool, and allows us to name the variable appropriately for both its type and its use.

lib/validation-lib.js
```diff
@@ -193,3 +196,27 @@ exports.jsonMatchesSchema = jsonMatchesSchema
...
+  }
+}
+
+exports.isResourceAdmin = isResourceAdmin
+
+exports.userHasProcessAccess = (device, process, userId, action) => (
+  or([
+    () => hasSufficientResourceAccess(process, userId, RESOURCE.PROCESS, action),
+    () => isResourceAdmin(device, userId),
+  ])()
```
DanielSeehausen: 👍

services/device/src/device/device.js
```diff
@@ -543,6 +555,28 @@ class Device {
   }
 
 
+  async getProcesses({ requestingUserId, deviceId }) {
+    const device = await this.devicesDB.findOne({ id: deviceId })
+    const user = await this.usersDB.findOne({ id: requestingUserId })
```
DanielSeehausen: (not for this pr, but for the future)
an opportunity for a combined lookup now that mo took the time to provide an example of how it can be done in mongo!

services/device/test/device/getProcesses.test.js
```diff
@@ -0,0 +1,82 @@
...
+      const { res, body } = await fireGetProcesses(device, notAdmin.id)
+      const processesIds = body.map(process => process.id)
+      const expectedProcessesIds = [process1.id, process2.id]
+      expectedProcessesIds.sort()
+      processesIds.sort()
+      expect(res.statusCode).toBe(200)
+      expect(processesIds.length).toBe(2)
+      expect(processesIds).toEqual(expectedProcessesIds)
+    })
+  })
```
DanielSeehausen: we'd be bad citizens if we only tested these two cases for a new endpoint
consider the different types of things being tested in `process/get.test.js` and think about recreating them here!


consider tests for other endpoints

services/process/package.json
```diff
@@ -2,7 +2,7 @@
   "scripts": {
     "seed-lib-files": "cp -rf ../../lib/* ./node_modules/",
     "pretest": "npm run seed-lib-files",
-    "test": "jest"
+    "test": "jest -i"
```
DanielSeehausen: same question re: if we need to go to in band here

services/process/src/dataPoints/dataPoints.js
```diff
@@ -11,15 +11,20 @@ const serializeStreamsData = require('../serializers/serializeStreamsData')
...
+    this.devicesDB = db.collection('devices')
     this.dataPointsDB = db.collection('dataPoints')
   }
 
   async get({ userId, processId, streamId, numberOfPoints, start, end }) {
     const process = await this.processesDB.findOne({ id: processId })
 
+    assert(() => canFind(process, `Process ${processId}`))
+
+    const device = await this.devicesDB.findOne({ id: process.deviceId })
```
DanielSeehausen: understanding this PR also introduced the ability to do this at once instead of in two calls the `process` model (which is accessible from `dataPoints`), we should make use of it!

DanielSeehausen: (what happens if the mongo query can't find the process, and therefore can't project in the device? does it return the same thing as an unfound process?)

services/process/src/process/process.js
```diff
@@ -38,7 +38,7 @@ class Process {
       comment: data.comment,
       ownerId: requestingUserId,
       deviceDescription: data.deviceDescription,
-      access:  createDefaultGroups('process', requestingUserId),
+      access: createDefaultGroups('process', requestingUserId),
```
DanielSeehausen: 👁 :laser:

services/process/src/process/process.js
```diff
@@ -95,65 +144,30 @@ class Process {
...
-      processIdsToDelete.map(processId => this.dataPointsDB.deleteMany({ processId }))
+  async deleteProcessesDataPoints(processIds) {
+    return Promise.all(
+      processIds.map(processId => this.dataPointsDB.deleteMany({ processId }))
     )
+  }
+
+  async deleteProcessesFromDevices(processesWithDevices) {
 
     const bulkDevices = this.devicesDB.initializeUnorderedBulkOp()
```
DanielSeehausen: Bjarne Stroustrup would recommend moving this variable declaration closer to its use. Currently, its declaration and its use sandwiches unrelated code to the variable!

services/process/src/process/process.js
```diff
@@ -95,65 +144,30 @@ class Process {
...
+    const processesIds = processesWithDevices.map(process => process.id)
 
-    targetDeviceIds.forEach(targetDeviceId => {
-      bulkDevices.find({ id: targetDeviceId }).updateOne({
-        $pull: { processes: { $in: processIdsToDelete }}
+    devicesIds.forEach(deviceId => {
+      bulkDevices.find({ id: deviceId }).updateOne({
+        $pull: { processes: { $in: processesIds } },
       })
     })
```
DanielSeehausen: is there a combination of `findAll` and `pullAll` we couldn't use here?

services/process/src/process/process.js
```diff
@@ -163,52 +177,66 @@ class Process {
 
     targetUserIds.forEach(targetUserId => {
       bulkUsers.find({ id: targetUserId }).updateOne({
-        $pull: { processes: { $in: processIdsToDelete }}
+        $pull: { processes: { $in: processesIds } },
       })
     })
```
DanielSeehausen: wondering again if there is a pullall combination we can use
we have an unfortunate reality here that users track their processes, and processes track their users. Not much we can do about that for the time being.
Unfortunately that leaves us open to a potentially bad state, where a user is no longer present in a processes 
access group, but they still have a reference to the process is on their user (or vice versa) :O

we may want to bite the bullet here and 
for all users, remove these processes ids. 

the potential slow speed won't be a problem until it is (which it certainly isn't now, and won't be anytime soon. Mongo is fast at such operations)

services/process/src/process/process.js
```diff
@@ -163,52 +177,66 @@ class Process {
...
+
+    processesWithDevices.forEach(processWithDevice => {
+      const { device } = processWithDevice
+      const { runningProcessId } = device
+
+      assertMany([
+        () => canFind(device, `Device ${processWithDevice.deviceId}`),
+        () => userHasProcessAccess(device, processWithDevice, requestingUserId, PROCESS_ACTIONS.DELETE),
+        () => hasSufficientResourceAccess(device, requestingUserId, RESOURCE.DEVICE, DEVICE_ACTIONS.OPERATE),
+        () => isNotPresent(
```
DanielSeehausen: we might want to update the joi schema to ensure the `processIdsToDelete` array can only contain uuidv4s. Otherwise, this has a false flag potential when checking `runningProcessId` (which should often be undefined or null), to match a bad value passed through in `processIdsToDelete`

services/process/src/process/process.js
```diff
@@ -163,52 +177,66 @@ class Process {
...
+          processIdsToDelete, runningProcessId, undefined, `Running Process with ID ${runningProcessId} can't be deleted`,
+        ),
+      ])
+    })
+
+    await this.deleteProcessesDataPoints(processIdsToDelete)
+
+    await this.deleteProcessesFromDevices(processesWithDevices)
+
+    await this.deleteProcessesFromUsers(processesWithDevices)
```
DanielSeehausen: very awesome abstraction work!

test/helper/Seeder.js
```diff
@@ -142,6 +139,10 @@ class Seeder {
     }
     const process = await this.seedProcess(user, device, processProps)
 
+    user.devices = [ deviceId ]
```
DanielSeehausen: should this be the responsibility of `seedUserDevice`?

test/helper/Seeder.js
```diff
@@ -142,6 +139,10 @@ class Seeder {
...
+    user.processes = [ processId ]
```
DanielSeehausen: and this the responsibility of `seedProcess` (which should maybe be named `seedUserProcess` for consistency, but this pr already huge so just ignore that :p)

test/helper/Seeder.js
```diff
@@ -142,6 +139,10 @@ class Seeder {
...
+    device.processes = [ processId ]
```
DanielSeehausen: this one would then belong here, as we have no `deviceProcess`!

services/process/src/process/process.js
vdiezel: There is a `updateMany`-call. Something like this COULD work: 

`await updateMany({ devicId: { $in: [ targetDeviceIds ] }}, { $pull: { processes: { $in: processIdsToDelete } } })`

I wrote bulk operations because I was a mongo noob when I wrote his.

services/process/src/process/process.js
vdiezel: The reasoning was that when I want to fetch e.g. all processes of a user, the only way to find all processes was to query the access groups of this process. I was afraid this will be a very slow operation, therefore this redundancy.

services/process/src/process/process.js
DanielSeehausen: pro

lib/validation-lib.js
```diff
@@ -198,9 +198,9 @@ exports.assertSchema = (schema) => (
 )
 
 const isResourceAdmin = (resource, userId) => {
-  const resourceUserGroup = resource.access.find(group => group.name === 'owners' && group.users.includes(userId))
+  const userIsOwner = resource.access.some(group => group.name === 'owners' && group.users.includes(userId))
```
DanielSeehausen: 👍

services/device/test/device/getProcesses.test.js
```diff
@@ -40,7 +40,7 @@ describe('getProcesses', () => {
     return { res, body: tryParseJSON(res.body) }
   }
 
-  describe('when user is admin', () => {
+  describe('when user is in owners group', () => {
```
DanielSeehausen: 👍 v nice

test/helper/generators/process.js
```diff
@@ -21,10 +21,17 @@ const traitDispatch = {
 
   CREATORASVIEWER: (processData, userId) => {
     const groups = processData.access
-    const operators = groups.find(group => group.name === 'viewers')
-    operators.users.push(userId)
+    const viewers = groups.find(group => group.name === 'viewers')
+    viewers.users.push(userId)
```
DanielSeehausen: good eye!

PR #204, opened 2020-01-22
202 break out trigger commands

### LINKED ISSUE #202: EVENTS: Break out trigger and command into new service
commands and triggers will be broken out into a new service

DanielSeehausen:
closes #202 

commits are broken up to be read linearly
will try to pair on review to speed through it


services/process/test/trigger/dataPointsToDB.test.js
```diff
@@ -3,8 +3,8 @@ const mongo = require('mongodb-lib')
 const uuidv4 = require('uuid/v4')
 
 const { promisifyHandler } = require('test-helpers/test-helper')
-const dataPointsToDbTrigger = promisifyHandler(require('../../src/iotTrigger/dataPointsToDb').main)
-const { createDataPointsToDbEvent } = require('./event-generators/createDataPointsToDbEvent')
+const dataPointsToDbTrigger = promisifyHandler(require('../../src/dataPoints/triggers/dataPointsToDB').main)
+const { createDataPointsToDbEvent } = require('./event-generators/createDataPointsToDBEvent')
```
vdiezel: `createDataPointstoDbEvent is being destructured while the name of the file has DB all in capitals. We should avoid confusion here and choose one.

DanielSeehausen: updated!

DanielSeehausen: good 👁

PR #205, opened 2020-01-22
203 create events trigger

DanielSeehausen:
best to review after #204 gets merged in, as this is branched off of it.
once that is merged, easiest to view commits in history.


vdiezel:
I approved #204 . Will review this once you have merged it.

services/process/src/dataPoints/triggers/dataPointsToDB.js.orig
```diff
@@ -0,0 +1,51 @@
...
+  const dataPoints = new DataPoints(db)
+
+  const result = await dataPoints.dataPointsToDB(requestData)
+
+  return success(result)
+}
+
+exports.main = middy(dataPointsToDB)
+  .use(logUnhandledErrors({ logger }))
+  .use(doNotWaitForEmptyEventLoop())
```
DanielSeehausen: ignore this! it was created by accident in the rebase, then deleted later

services/device/test/trigger/createEvent.test.js
```diff
@@ -0,0 +1,136 @@
...
+          await fireCreateEventEvent(payloadA)
+          await fireCreateEventEvent(payloadB)
+
+          const eventsCount = await dbSeeder.eventsDB.count()
+          expect(eventsCount).toBe(0)
+        })
+      })
+    })
+  })
+})
```
DanielSeehausen: any obvious tests I am missing?

services/device/src/device/device.js
```diff
@@ -497,6 +498,21 @@ class Device {
   /*********************************************************************
                                 TRIGGERABLES
   **********************************************************************/
```
DanielSeehausen: image
image
image

DanielSeehausen: hope that helps clarify the pr

test/helper/Seeder.js
```diff
@@ -29,10 +39,9 @@ class Seeder {
   }
 
   async flushDatabase() {
-    await this.usersDB.deleteMany({})
-    await this.devicesDB.deleteMany({})
-    await this.processesDB.deleteMany({})
-    await this.dataPointsDB.deleteMany({})
+    Promise.all(this.collections.map((collection) => (
+      collection.deleteMany({})
+    )))
```
vdiezel: Don't you have to `await` here or return the promise so it's awaited by the function call? I really need to figure this out ...

services/device/src/device/triggers/createEvent.js
```diff
@@ -0,0 +1,39 @@
...
+  const db = await mongo.getDB()
+  const device = new Device(db)
+
+  const result = await device.createEvent(requestData)
+  return success(result)
+}
+
+exports.main = middy(createEvent)
+  .use(logUnhandledErrors({ logger }))
+  .use(httpErrorHandler({ logger }))
```
vdiezel: `httpErrorHandler` is not necessary here because it's not a HTTP request.

services/device/test/trigger/createEvent.test.js
```diff
@@ -0,0 +1,136 @@
...
+      test('successfully creates the event', async () => {
+        const payload = getValidPayload()
+        await fireCreateEventEvent(payload)
+
+        const allEvents = await dbSeeder.eventsDB.find().toArray()
+
+        expect(allEvents.length).toBe(1)
+
+        const firstEvent = allEvents[0]
+        expect(firstEvent).toHaveProperty('_id')
```
vdiezel: :+1:

services/device/serverless.yml
```diff
@@ -624,6 +624,19 @@ functions:
...
 ################### IoT Triggers ##################
 ###################################################
+  createEventTrigger:
+    role: AWSLambdaLoggingRole
+    handler: src/device/triggers/createEvent.main
+    events:
+      - iot:
+          name: createEvent_${{self:service}}_${{self:provider.stage}}
+          sqlVersion: '2016-03-23'
+          sql: "SELECT *, topic(4) AS deviceId FROM '${{self:provider.stage}}/event/create/+'"
```
vdiezel: looks good. We have to add this topic to the device policy for all existing devices and also add it to the creation of the device policy in libs/utils.

vdiezel: Would it make more sense to switch `deviceId (+)` with `create` in the topic?

test/helper/Seeder.js
DanielSeehausen: good eye! Yes, I definitely need to return the Promise.all
to follow up, the caller can only act on the return value, which is undefined. 

We can omit the `async` on `flushDatabase` while returning the promise.all, but I prefer to keep it (the `async`) as its a good communicator that this function returns a promise, and can be awaited where it is called. How do you feel about that?

services/device/serverless.yml
DanielSeehausen: updated url to:
`'${{self:provider.stage}}/device/+/event/create'`

services/device/src/device/device.js
```diff
@@ -497,6 +498,21 @@ class Device {
...
+  async createEvent({ deviceId, eventType, name, createdAt, processId, displayMsg }) {
+
+    const event = {
+      deviceId,
```
vdiezel: I think the event should also get a unique ID which is also a uuidv4. Currently we could rely on the `_id` provided by MongoDB but we broke with that pattern last year to have it uniform, lets stick to it.

DanielSeehausen: good point - will update

services/device/test/trigger/createEvent.test.js
vdiezel: You could additionally - as the cherry on top - check if the correct unhandled error is thrown by the middleware when the validation fails :japanese_ogre:.

test/helper/Seeder.js
vdiezel: agreed.

services/device/serverless.yml
DanielSeehausen: this has been updated! see utils

services/device/src/device/device.js
DanielSeehausen: updated - see new helper function for spec

services/device/test/trigger/createEvent.test.js
DanielSeehausen: i was under the impression mqtt got some sort of special response, or destroyed the object
i'd be happy to add that if that's not the case

services/device/src/device/triggers/createEvent.js
DanielSeehausen: to follow up with your above comment, I meant to take this out. I had it in here to try to get validations to return useful info, then I remembered something about return values for mqtt being butchered or something. Ill remove and save an mqtt error handler for another time to catch validation response strings.

DanielSeehausen:
utils tests throwing - fixed

services/device/test/trigger/createEvent.test.js
DanielSeehausen: going to save this refactor for another time as it would involve making an mqtt error handler middleware if we were doing it right, and then we could apply that to more functions

services/device/src/device/triggers/createEvent.js
DanielSeehausen: removed

DanielSeehausen: oops! I was wrong in this
i had it in there because it wraps up and handles the validation lib throwing errors. Unless we made an mqtt specific one, it would throw an error and not return properly. I am going to leave in for now for this reason. At some point, this warrants a refactor.

DanielSeehausen:
@vdiezel this should be good to go now. see comment on `httpErrorHandler`

services/device/test/trigger/createEvent.test.js
vdiezel: I meant when the validation in the lambda handler fails, Joi will throw an error that should be caught by our ''logUnhandledErrors" middleware.

PR #206, opened 2020-01-23
finished log service creation

### LINKED ISSUE #201: create a new log service on hambda
Implement a new service to handle logs

vdiezel:
creates a log service which exposes a create log function that can be triggered via SNS. 

- implements a SNSLogAPI in libs to be used from other services to create logs
- implements  a new API gateway and adds access to it to the authporized cognito role in the user service, __it means user service needs to be redeployed before HTTP requests to the log service will work__
- implements a new policy that has to be attached to the invocation role of every lambda function that wants to create logs. Attached it already to the two default roles in `process` and `device`
- implements a new middleware to parse payload from SNS messages 

Functions that need to create logs need to have the `STAGE` env var defined in the serverless.yml file.

Current implementation relies on MongoDB for testing purposes. Which storage is going to be used finally is not decided yet.

`env` files need to be updated! Added the API gateway ID to the ones in pinned files on slack.

closes #201 

### Update 

- adds caching of all services in the circleCI config
- removes dynamodb-lib from libs


lib/middleware/parseSNSMessagePayload.js
```diff
@@ -0,0 +1,25 @@
...
+  } else {
+    handler.event.messagePayload = tryJsonParse(payload)
+  }
+
+  return next()
+}
+
+module.exports = ({ parseToJSON = true } = {}) => ({
+  before: (handler, next) => middleware(handler, next, parseToJSON),
+})
```
vdiezel: message payload of SNS is a string, so logs arrive as stringified JSONs. This middleware tries to parse them out.

services/device/serverless.yml
```diff
@@ -92,6 +92,20 @@ resources:
...
+                - Effect: Allow
+                  Action:
+                    - sns:Publish
+                  Resource:
+                    - 'Fn::Join':
+                        - ':'
+                        - - 'arn:aws:sns'
+                          - Ref: 'AWS::Region'
+                          - Ref: 'AWS::AccountId'
+                          - '${{self:provider.stage}}-createLog'
```
vdiezel: This one was tricky to set up: using variables within `Fn::Join` requires the variable to be wrapped in quotes.

services/log/src/log/list.js
```diff
@@ -0,0 +1,43 @@
...
+
+exports.main = middy(list)
+  .use(logUnhandledErrors())
+  .use(httpErrorHandler({ logger }))
+  .use(doNotWaitForEmptyEventLoop())
+  .use(setRequestingUserId())
+  .use(httpEventNormalizer())
+  .use(jsonBodyParser())
+  .use(stringifyResponseBody())
+  .use(cors())
```
vdiezel: This function doesn't do anything yet.

services/user/serverless.yml
```diff
@@ -154,6 +155,10 @@ resources:
                   Action:
                     - 'execute-api:Invoke'
                   Resource: arn:aws:execute-api:${{self:provider.region}}:*:${{self:provider.utilsApiGatewayId}}/*
+                - Effect: 'Allow'
+                  Action:
+                    - 'execute-api:Invoke'
+                  Resource: arn:aws:execute-api:${{self:provider.region}}:*:${{self:provider.logApiGatewayId}}/*
     CognitoUnauthRole:
       Type: AWS::IAM::Role
       Properties:
```
vdiezel: Adds access to the API gateway of the authorized user role.

lib/request-lib.js
```diff
@@ -0,0 +1 @@
+exports.getSNSMessagePayload = event => event.messagePayload
```
vdiezel: messagePayload on the event is set via the new middleware. Wanted to decouple how it is retrieved from the lambda handler. Maybe a bit overengineered, but if we modify the middleware, we don't have to modify all the handlers, just this function.

lib/middleware/parseSNSMessagePayload.js
```diff
@@ -0,0 +1,25 @@
+const tryJsonParse = string => {
+  try {
+    const parsedContent = JSON.parse(string)
+    return parsedContent
+  } catch (_) {
+    return string
+  }
+}
```
DanielSeehausen: same thing same name is created in 
`test-helper.js`

makes sense you don't want to to import it being in test-helper, just mentioning. nice to have the middleware encapsulated

lib/request-lib.js
DanielSeehausen: 👍 love it

lib/middleware/parseSNSMessagePayload.js
DanielSeehausen: any way we could sneak `JSON` into the title of the middleware, so the caller knows what it is attempting to parse to?

services/log/serverless.yml
```diff
@@ -0,0 +1,87 @@
...
+                    - logs:CreateLogStream
+                    - logs:PutLogEvents
+                    - logs:CreateLogGroup
+                  Resource: '*'
+###################################################
+################### Lambda ########################
+###################################################
+functions:
+  createLog:
+    handler: src/log/create.main
```
DanielSeehausen: because this is a new service, I thought i'd ask now:
is `.main` required or could we return a default export?

services/log/serverless.yml
```diff
@@ -0,0 +1,87 @@
...
+    role: AWSLambdaLoggingRole
+    environment:
+      DB_USER: ${{file(../../env.${{self:provider.stage}}.yml):DB_USER}}
+      DB_PASSWORD: ${{file(../../env.${{self:provider.stage}}.yml):DB_PASSWORD}}
+      DB_NAME: ${{file(../../env.${{self:provider.stage}}.yml):DB_NAME}}
+    events:
+      - sns: ${{self:provider.stage}}-createLog
+  listLogs:
+    handler: src/log/list.main
```
DanielSeehausen: also because this is a new service, wanted to make a shameless plug for RESTful routing/action names

[`index` is the action for GET all of a resource](https://routes.readthedocs.io/en/latest/restful.html)
it breaks out convention, so probably not worth it right now, but wanted to point it out.

lib/middleware/parseSNSMessagePayload.js
vdiezel: Because I provided an option it doesn't always parse to JSON. That is why I ommited it.

services/log/serverless.yml
vdiezel: It doesn't have to be a 'main', you can use any name here, in the docs you will often find 'handler' or 'main'. But I don't know if we could use a default export instead of a named one. I think we tried default export when implementing middy and it didn't work.

services/log/serverless.yml
vdiezel: I stayed with list for the simple reason that having a file named `index.js` for this handler is confusing :japanese_goblin:

.circleci/config.yml
```diff
@@ -13,12 +13,31 @@ jobs:
     steps:
       - checkout
 
-        # having a hard time restoring multiple node_module caches, so I just did device for now since it should be the largest. This seems to be a bug that is reported multiple times in circleCI forums, but I have seen no real solution to it yet...The problem shows itself when trying to restore multiple services' caches -- circleCI simply restores the first `restore_cache` called in this script, then it skips the rest and goes straight to `npm install` :(  ds 11/11/19
```
DanielSeehausen: I remember this now
seems this implementation either fixed it or it was resolved, as circle ci is logging success

image

lib/middleware/parseSNSMessagePayload.js
DanielSeehausen: 👍

services/log/serverless.yml
DanielSeehausen: believe I recall the same

PR #215, opened 2020-01-24
214 index logs endpoint

### LINKED ISSUE #214: Implementing fetching logs into hambda logs service
Provide a GET endpoint to fetch logs for a device; finalize log creation and provide test coverage

vdiezel:
first three commits are in the previous PR #206 . 

- finishes creation of logs in MongoDB
- implements test for both functions `create` and `list`

closes #214


lib/request-lib.js
```diff
@@ -0,0 +1,13 @@
...
+/* helper function to return undefined when parsing query
+ * string parameters to numbers because parseInt returns NaN if they are not set
+ * and NaN overrides default parameters set in the function
+ */
+
+const parseQueryStringParameterToNumber = string => {
+  const parsedNumber = parseInt(string, 10)
+  if (!Number.isNaN(parsedNumber)) return parsedNumber
+}
+exports.parseQueryStringParameterToNumber = parseQueryStringParameterToNumber
```
vdiezel: This function helps parsing out queryStringParameters to numbers. `parseInt` returns NaN when it can't parse the input. In JS NaN does overwrite default parameters, so this function is a wrapper thats returns `undefined` instead of NaN which guarantees that default parameters defined in the handler are applied.

services/log/src/log/log.js
```diff
@@ -0,0 +1,83 @@
...
+      .find(query)
+      .sort({ createdAt: 1 })
+      .limit(limit)
+      .toArray()
+
+    return logs
+  }
+}
+
+module.exports = Log
```
vdiezel: This is what I came up with. Logs are fetched using time ranges. When no time ranges are provided, I simply fetch the most recent ones. You can also provide a limit and a processId. 

On the frontend, on intial load you load the most recent logs. I will provide two options to reload logs, one will load more old logs by simply using the startTime of the last laoded log as the endTime in the request. 
When I want to check for more newly created logs, I will simply send the endTime of the last log as a startTime and not provide any endTime. Thinking of 2 buttons (or on scroll), one will be `update` for recent logs and `load more` for older logs. This way I don't have to load logs from the database, compare to what the user already has and load missing logs to synchronize with the user.

vdiezel: The authorization flow needs a rework as I just realized. For every log a user tries to load we want to check if that log has a processId associated with it and then we would need to check if the user has access to this process. Or we move checking logs to such a high authorization level that we don't have to do this.

services/log/test/log/create.test.js
```diff
@@ -0,0 +1,168 @@
...
+      expect(fetchedLogs.length).toBe(1)
+
+      const firstLog = fetchedLogs[0]
+      expect(firstLog).toMatchObject(newLogEntry)
+    })
+  })
+
+  describe('when creating a log with invalid data', () => {
+
+    // wrapping call into try as a workaround before Joi validation is protocol agnostic
```
vdiezel: :bangbang:

services/log/test/log/create.test.js
```diff
@@ -0,0 +1,168 @@
...
+
+      const logCount = await dbSeeder.logsDB.count()
+      expect(logCount).toBe(0)
+    })
+
+    test('does not create a new log for a wrong level', async () => {
+      const newLogEntry = generateLog()
+      newLogEntry.level = 'Insecure about what it wants to be'
+
+      const errorMsg = 'must be one of [information, warning, error]'
```
vdiezel: Added a bit more context to the validations so test are more self-descriptive

lib/SNSLogAPI.js
```diff
@@ -0,0 +1,33 @@
...
+ * Log Model:
+ *
+ * deviceId  uuidv4 (required)
+ * processId uuidv4 (optional)
+ * level: 'information' | 'warning' | 'error' String (required)
+ * message: String (length: 0 - 256) (required)
+ * subject: String (length: 0 - 64) (optional)
+ * createdBy: uuidv4 (required) - origin of the log (device, user - maybe process?)
+ *
+ */
```
DanielSeehausen: this is helpful, but a recreation of the joi schema. If the schema changes, then we will have to find this too and change it. If it is to make sure devs aren't publishing logs they shouldn't, consider referencing the joi file instead.

lib/request-lib.js
DanielSeehausen: This would parse to int instead of number :(
'34.44' --> 34

This would also go through:
'24HELLO' --> 24

just to cheat on checking corner cases because stuff like this is always riddled with little gotchas, I looked at [yerboi sindresorhus' implementation](https://github.com/sindresorhus/query-string/blob/master/index.js#L185)

Here is their implementation:

```js
if (!Number.isNaN(Number(value)) && (typeof value === 'string' && value.trim() !== '')) {
  value = Number(value);
```

services/log/src/log/create.js
```diff
@@ -0,0 +1,28 @@
...
+async function create(event) {
+
+  const requestData = getSNSMessagePayload(event)
+  validateCreate(requestData)
+
+  const db = await mongo.getDB()
+  const log = new Log(db)
+
+  await log.create(requestData)
+  return 1
```
DanielSeehausen: any objections to tying the return value to whether the log was created successfully or not? Does this return value go back to the caller through the sns service?

services/log/src/log/log.js
```diff
@@ -0,0 +1,102 @@
...
+
+  async create(logData) {
+    const logEntry = Log.createLogEntry(logData)
+    await this.logsDB.insertOne(logEntry)
+  }
+
+  static normalizeProcessId(log) {
+    const processId = log.processId === null ? undefined : log.processId
+    return { ...log, processId }
+  }
```
DanielSeehausen: is this because there will be logs stored in the database that can have either `null` or `undefined`?
Any reason we don't want to do this on the other end (before the log gets persisted, we serialize to undefined or null)?

is there a reason we choose undefined over null, given that null is more specific?

services/log/src/log/log.js
```diff
@@ -0,0 +1,102 @@
...
+
+  async createAuthorizedLogQuery(device, user, processId) {
+
+    const userIsAdmin = device
+      .access.some(group => group.name === 'owners' && group.users.includes(user.id))
```
DanielSeehausen: is there a validator that we could use for this? or some type of access method? 
If not, any chance we could add admins (as in users who are admins) to this check? Do we want people with 'developer' level to have access to these logs?

services/log/src/log/log.js
```diff
@@ -0,0 +1,102 @@
...
+    // using Mohammeds approach
+
+    if (userIsAdmin) return { processId }
+
+    assert(() => hasSufficientResourceAccess(
+      process, user.id, RESOURCE.PROCESS, PROCESS_ACTIONS.GET,
+    ))
+
+    return { processId }
+  }
```
DanielSeehausen: I see what you mean about how this can get hairy. I have a question but ill come ask you in person

services/log/src/joi/log.js
```diff
@@ -0,0 +1,32 @@
+const Joi = require('@hapi/joi')
+const { assertSchema } = require('validation-lib')
+
+const UUIDV4_REGEX = /[a-f0-9]{8}(-[a-f0-9]{4}){3}-[a-f0-9]{12}/
+
+const VALID_LOG_LEVELS = ['information', 'warning', 'error']
+
+const create = assertSchema(Joi.object({
+  deviceId: Joi.string().required().pattern(UUIDV4_REGEX),
+  processId: Joi.string().pattern(UUIDV4_REGEX),
```
DanielSeehausen: if this is not present could we set this to `null` through joi? would that make sense, or do we need `undefined`?

lib/request-lib.js
DanielSeehausen: had I read the tests first, I would see that this was your desired behavior re: floats to ints and int led strings to ints. Is that intentional? If so, could I put in a request to rename this `...ToInt` to avoid ambiguity

vdiezel: I renamed the function. The edge case covered is when an empty string is passed into `Number()` coercion, it returns 0. `parseInt` doesn't do that. Number has another advantage that it removes trailling zeros for float numbers (which e.g. is used in our number utils file on the frontend).

lib/SNSLogAPI.js
vdiezel: done.

services/log/src/log/log.js
vdiezel: MongoDB always stores `undefined` as `null`. But passing `nul`l as a value to our handler confuses Joi because `null` overwrites defaults and is considered by Joi as a set property and will apply validation on top. It doesn't when it is `undefined`. I think we can have both, I just need to fix the test.

services/log/src/log/log.js
vdiezel: This will be modified once we have the new validators Mo is working on.

services/log/src/joi/log.js
vdiezel: We now take undefined, and when it is dumped into MongoDB it becomes null and is also fetched as null from the DB.

lib/request-lib.js
```diff
@@ -1,13 +1,13 @@
...
-const parseQueryStringParameterToInt = string => {
-  const parsedNumber = parseInt(string, 10)
-  if (!Number.isNaN(parsedNumber)) return parsedNumber
+const tryParsingToNumber = value => {
+  const parsedNumber = Number(value)
+  if (Number.isNaN(parsedNumber) || (typeof value === 'string' && value.trim() === '')) return value
+  return parsedNumber
 }
-exports.parseQueryStringParameterToInt = parseQueryStringParameterToInt
+exports.tryParsingToNumber = tryParsingToNumber
```
DanielSeehausen: just learned:
"Casting : is conversion of an object from a similar datatype; like from int to double or from decimal to int. ... Parsing : Parsing, on the other side, is converting from one data type to another, like from string to int for example"

services/log/src/log/log.js
DanielSeehausen: :pro:

DanielSeehausen: just need to remember to update it - perhaps ping mo or sit with him after he rebases to add it? not sure what best way to do that would be.
or wait to rebase on mo's

PR #216, opened 2020-01-24
209 create alert limit api

### LINKED ISSUE #209: 1 alert: Create alert emit API
Create alert emit API

DanielSeehausen:
(provided inline comments - should help read through)

- makes the event model top level so it can be shared (see its use in create device event), and takes us another step in a functional direction with `Connectable` class
- Device iot publishes moved outside of the `Device` model into the calling lambda functions for `setCalibrarion`, `deleteCalibration`, and `sendDeviceDescription`. With that, any information required to make the decision of "should I not publish something over iot/mqtt" has also been returned to the lambda so it can make the decision to follow through and make the mqtt publish
- creates new mqtt message publisher for events in `IotDataAPI`
- there was some inconsistency around seeding userDevice/userAndDevice/device (without user). Mostly old tech debt from me being in a rush and lazy with consistency/naming. In order to simplify things and increase transparency, (and most importantly to provide an option to seed a device without a user in a way that things wont break, as we were before seeding userAndDevice and trying to pass in non-breaking values for a user), a simple `seedDevice` has been created, and `seedUserDevice` has been matched up with `generateUserDevice` to both expect an actual user, that a device is being seeded for. 
- tests that were seeding userDevice in a variety of ways have been standardized/fixed (it might be we were seeing user and device quite often while we only needed a device - because it would break otherwise - but I didn't go through and check)
- `count` -> `countDocuments`, as the former is getting deprecated 

I'd love feedback on the general concept around bringing the model closer to something truly functional. Particularly in regards to the `Connectable` class usage, or some concept like that.

closes #209


lib/IotDataAPI.js
```diff
@@ -1,16 +1,19 @@
 const AWS = require('aws-sdk')
 const { badRequest, iotUnauthorized, iotMethodNotAllowed, failure } = require('./response-lib')
 
+AWS.config.update({ region: process.env.OSPIN_AWS_REGION })
```
DanielSeehausen: moved this out of the execute function, as config stuff is set on initial setup/file requiring. If it is getting edited during running, we are doing something very funky. Eventually, this shouldn't happen here, but its ok to see it explicitly for now until we have time to refactor!

lib/IotDataAPI.js
```diff
@@ -22,6 +25,20 @@ class IotDataAPI {
...
+      payload: event,
+    }
+
+    return this.execute(mqttMsg)
+  }
+
+  static publishDeviceEvent(deviceId, event) {
+    return this.publishResourceEvent(deviceId, 'device', event)
+  }
+
```
DanielSeehausen: here you have the generic `publishResourceEvent`, as well as the specific `publishDeviceEvent` which makes use of it. This leaves room for things like `publishUserEvent` in the future.

lib/dynamodb-lib.js
```diff
@@ -1,33 +0,0 @@
...
-          setTimeout(checkStatus, 200);
-        }
-      } catch (e) {
-        reject(e);
-      }
-    };
-
-    setTimeout(checkStatus, 200);
-  });
-};
```
DanielSeehausen: this seemed to be unused!

lib/models/Connectable.js
```diff
@@ -0,0 +1,15 @@
...
+    this.eventsDB = this.db.collection('events')
+    this.devicesDB = this.db.collection('devices')
+    this.usersDB = this.db.collection('users')
+    this.processesDB = this.db.collection('processes')
+    this.dataPointsDB = this.db.collection('dataPoints')
+  }
+
+}
+
+module.exports = Connectable
```
DanielSeehausen: this class is intended to be extended by the models to provide a unified interface to the database. Its potential use case would be to replace calls such as:
`const device = await new Device(db)`
by:
`await Device.connect(db)`

it would remove the duplicate constructors in our models as well. See `Event` class for an example of how it can be extended and used

DanielSeehausen: One consideration I struggled with was:
should `connect` actually async the `getDB` to ensure the db is started up?

the plus of that being:
- we won't need to await the DB connection always before we call connect. 
the negative being:
- `Connectable` starts importing the DB library, then which DB to use becomes baked into `Connectable`. 

I thought the negative was greater than the positive for now, and figured it would be easier to build on simply passing the DB for the time being.

lib/models/Event.js
```diff
@@ -0,0 +1,49 @@
...
+    const eventTypeSupported = this.EVENT_TYPES.has(eventType)
+    if (eventTypeSupported) return eventType
+
+    console.warn(`${eventType} is not a supported event type!`)
+    return 'notification'
+  }
+
+  static isAlarm({ eventType }) {
+    return eventType && eventType === 'alarm'
+  }
```
DanielSeehausen: not all events are considered alarms. See where this is being used in devices `createEvent` lambda

lib/models/Event.js
```diff
@@ -0,0 +1,49 @@
...
+      createdAt,
+      processId,
+      displayMsg,
+    }
+
+    const result = await this.eventsDB.insertOne(event)
+    const success = result.insertedCount === 1
+
+    return { created: success, event }
+  }
```
DanielSeehausen: we expect later one that `deviceId` won't be required, so this should be easy to change.

lib/models/Event.js
```diff
@@ -0,0 +1,49 @@
...
+      id: uuidv4(),
+      deviceId,
+      eventType: parsedEventType,
+      name,
+      createdAt,
+      processId,
+      displayMsg,
+    }
+
+    const result = await this.eventsDB.insertOne(event)
```
DanielSeehausen: here you see the extended class `Connectable` being used with the `super.eventsDB`

services/device/src/device/deleteCalibration.js
```diff
@@ -23,16 +24,26 @@ async function deleteCalibration(event) {
...
+  const calibrationDeleted = await device.deleteCalibration(requestData)
+
+  if (!calibrationDeleted) return success(0)
+
+  await IotDataAPI.publishCommand(requestData.deviceId, {
+    command: 'calibrate',
+    streamId: parseInt(requestData.streamId, 10),
+    slope: 1,
+    offset: 0,
+  })
```
DanielSeehausen: this was a first step in moving Iot calls outside of the models themselves.

services/device/src/device/device.js
```diff
@@ -1,20 +1,20 @@
 const uuidv4 = require('uuid/v4')
 
 const { assert, assertMany, canFind, hasDeveloperAccess, hasSufficientResourceAccess, isNotPresent, streamValueWithinLimits, processNotStartedYet, isMyResource, hasResourceAccess } = require('validation-lib')
-const IotDataAPI = require('IotDataAPI')
```
DanielSeehausen: no longer required to keep the model ignorant of these side effects

services/device/src/device/device.js
```diff
@@ -1,20 +1,20 @@
...
 const IotAPI = require('IotAPI')
 const { DEVICE_ACTIONS, PROCESS_ACTIONS, createDefaultGroups, RESOURCE } = require('access-rights-lib')
+const Event = require('models/Event')
```
DanielSeehausen: because this was about making an endpoint for `CREATE device/<id>/event`, the device is going to add what information is important for a device event to a normal event payload then make use of the event model (right now, just the deviceId).

services/device/src/device/device.js
```diff
@@ -354,50 +353,30 @@ class Device {
...
-        command: 'calibrate',
-        streamId: parseInt(data.streamId, 10),
-        slope: data.params.slope,
-        offset: data.params.offset
-      }
-
-      await IotDataAPI.publishCommand(deviceId, payload)
-    }
-
-    return result.modifiedCount
```
DanielSeehausen: iot call placed into calling function

services/device/src/device/device.js
```diff
@@ -354,50 +353,30 @@ class Device {
...
-
-    if (device.streams.some(stream => stream.id === streamId)) {
-      const payload = {
-        command: 'calibrate',
-        streamId: parseInt(streamId, 10),
-        slope: 1,
-        offset: 0
-      }
+    const calibrationExists = device.streams.some((stream) => stream.id === streamId)
+    if (!calibrationExists) return false
```
DanielSeehausen:
```diff
! would you mind taking a look at this part in particular
```
considering this now exits early if it can't find the stream requested, are there any potential problems? 

It looked like previously we attempted to update the device by pulling the calibration of a specific streamId, _then_ went to check if the device actually had the stream.

services/device/src/device/device.js
```diff
@@ -499,9 +478,10 @@ class Device {
                                 TRIGGERABLES
   **********************************************************************/
   async createEvent({ deviceId, eventType, name, createdAt, processId, displayMsg }) {
+    const device = await this.devicesDB.findOne({ id: deviceId })
+    assert(() => canFind(device, `Device ${deviceId}`))
```
DanielSeehausen: don't allow creation of device events for devices that can not be found

services/device/src/device/setCalibration.js
```diff
@@ -22,16 +23,26 @@ async function setCalibration(event) {
...
+  if (!calibrationSet) return success(0)
+
+  await IotDataAPI.publishCommand(requestData.deviceId, {
+    command: 'calibrate',
+    streamId: parseInt(requestData.data.streamId, 10),
+    slope: requestData.data.params.slope,
+    offset: requestData.data.params.offset,
+  })
+
+  return success(1)
```
DanielSeehausen: here is an example of how the iot action is abstracted from the model method.

services/device/src/device/triggers/createEvent.js
```diff
@@ -14,23 +16,31 @@ const validateCreateEvent = require('../../joi/triggers').createEvent
...
 
-  const result = await device.createEvent(requestData)
-  return success(result)
+  const { created, event } = await device.createEvent(eventData)
+
+  if (!created) return success(0)
+
+  if (Event.isAlarm(event)) {
+    IotDataAPI.publishDeviceEvent(event.deviceId, event)
+  }
```
DanielSeehausen: circling back on the Event model to determine if this event data is considered alarm worthy

services/device/test/trigger/createEvent.test.js
```diff
@@ -2,14 +2,16 @@ const uuidv4 = require('uuid/v4')
 const { MongoClient } = require('mongodb')
```
DanielSeehausen: these tests are a little hard tor ead with the diff. Maybe just easier running them plainly or reading through the non-diff file's describe/test quotes

services/device/test/trigger/createEvent.test.js
```diff
@@ -51,86 +53,148 @@ describe('createEvent trigger', () => {
...
+            await fireCreateEventEvent(payload)
 
-          await fireCreateEventEvent(payload)
-          const eventsCount = await dbSeeder.eventsDB.count()
+            const event = (await dbSeeder.eventsDB.find().toArray())[0]
 
-          expect(eventsCount).toBe(0)
+            expect(IotDataAPI.publishDeviceEvent)
+              .toHaveBeenCalledWith(payload.deviceId, event)
+          })
```
DanielSeehausen: this the 🍖

services/device/src/device/device.js
vdiezel: A user could update on a stream that has been disconnected ( e.g. on an old process). We still want to save the calibration then, but we don't need to publish message.

services/device/src/device/device.js
vdiezel: same here as with delete calibration

vdiezel: Scratch what I wrote before here. This one can stay as it is. We don't allow calibrating on something disconnected. Line 327 of this file was preexistent and manages that. Only deleting calibration needs a fix.

PR #223, opened 2020-02-04
211 populate users with their events

### LINKED ISSUE #211: 5 alert: populate users with their events
so that when devices are fetched, so are alerts

DanielSeehausen:
### Update 1
[up to and including the 4th commit, starting here](https://github.com/ospin-web-dev/hambda/pull/223/commits/f3084bb5d67079a079f125eac1d0b3add616cef8)

These commits cover the creation of the `mixin` helper, which is a composer for class static methods. Eat your heart out lodash, I hope you like it. Highly recommend reading the tests - they should document what it is supposed to do, and how it is being used, pretty well.

---
### Update 2
[including the 5th to the 11th commit, starting here](https://github.com/ospin-web-dev/hambda/pull/223/commits/b02a3f1a9bdc41be0f657caf2bfbbdd9f5238295)

This moves the seeder to be functional and updates references to it. It removes a lot of duplicate work, duplicate ways to interface with the db, erroneous `connection` object creation and closing, which was causing the `open handlers` issue which hung the tests.

This will look like a mess reading plainly. Viewing the commits should be pretty streamlined though, and make the changes which effected many files in the exact same way fast to read.

---
### Update 3
(up until here, everything was refactoring and not changing the API)
[including the 5 commits starting here](https://github.com/ospin-web-dev/hambda/pull/223/commits/f586ac445a7ee1354e611a5fd91f3efda84a7e96)

This is the actual deliverable of this PR. Not much to see:
- adds a user find query that populates in the user/devices/events into the user object
- updates GET `user/<id>` to make use of the query
- completes the test coverage for the the query

---

### Post edit update
- Made requested changes
- moved seeder stuff to libs as it was causing difficult to track issues with `seed-lib-files` and belongs in shared libs
- updated assertSchema to createSchemaAssertion to more accurately reflect what the method is doing
---
closes #211


test/helper/Seeder.js
```diff
@@ -1,37 +1,22 @@
...
-    this.collections = [
-      this.devicesDB,
-      this.processesDB,
-      this.usersDB,
-      this.dataPointsDB,
-      this.eventsDB,
-      this.logsDB,
-    ]
+  static async useDB(db) {
+    this.connect(db)
```
DanielSeehausen: this line could be removed in an ideal situation in which the seeder is always engaging the DB through the models (like it is for `Event` currently). For now, it must remain though as the collections are references all throughout seeder.

test/helper/generators/event.js
```diff
@@ -0,0 +1,17 @@
...
+const generateEvent = (data) => (
+  Event.new({
+    deviceId: uuidv4(),
+    eventType: 'notification',
+    name: faker.hacker.noun(),
+    createdAt: Date.now(),
+    processId: null,
+    displayMsg: faker.hacker.phrase(),
+    ...data,
+  })
```
DanielSeehausen: calls on `Event`'s new, which is responsible for enforcing the schema
`Event.create`'s responsibility becomes the actual persisting of the record.

lib/functional-helpers/mixin.js
```diff
@@ -0,0 +1,65 @@
...
+    keys = keys.concat(getOwnPropertySymbols(object))
+  }
+
+  return keys
+}
+
+function mixinStatics(targetClass, mixableClass) {
+
+  if (typeof targetClass !== 'function' || typeof mixableClass !== 'function') {
+    return targetClass
```
vdiezel: Should we throw a warning here? I can see how a bug caused by this would be tricky to find.

lib/models/Connectable.js
```diff
@@ -10,6 +14,20 @@ class Connectable {
     this.usersDB = this.db.collection('users')
     this.processesDB = this.db.collection('processes')
     this.dataPointsDB = this.db.collection('dataPoints')
+    this.logsDB = this.db.collection('logs')
+  }
+
+  static get collections() {
+    if (!this.db) return []
```
vdiezel: `if (!this.isConnected) return []`

lib/models/Event.js
```diff
@@ -26,26 +26,31 @@ class Event {
     return eventType && eventType === 'alarm'
   }
 
-
-  static async create({ deviceId, eventType, name, createdAt, processId, displayMsg }) {
-    const parsedEventType = Event.parseEventType(eventType)
-
-    const event = {
+  static new({ deviceId, eventType, name, createdAt, processId, displayMsg }) {
```
vdiezel: It could be confusing using `new` here because the first intuition of a dev would be that it creates a new instance of a class by invoking its constructor. But use your judgement here. Maybe `generate` ( even though we use this name for the Seeder). `createEventObject`, `createEventRecord`. If nothing convincing comes to mind, lets go with new.

test/helper/generators/event.js
vdiezel: :fire:

lib/functional-helpers/mixin.js
```diff
@@ -0,0 +1,65 @@
...
+
+  try { // Avoid failures from read-only properties
+    defineProperty(receiver, key, descriptor)
+  } catch (e) { /* silently ignore */ }
+}
+
+function getObjectKeys(object) {
+  let keys = getOwnPropertyNames(object)
+
+  if (getOwnPropertySymbols) {
```
vdiezel: maybe do a quick research if this is necessary in NodeJS

services/user/test/queries/getWithEvents.test.js
```diff
@@ -25,19 +25,62 @@ describe('User\'s getWithEvents query', () => {
...
+  describe('when the user is an admin on the event devices', () => {
+    const deviceTraits = [ 'USER_AS_ADMIN']
 
-    const alarm = await Seeder.seedDeviceEvent(device, { eventType: 'alarm' })
-    const notification = await Seeder.seedDeviceEvent(device, { eventType: 'notification' })
+    test('loads a user with all of their device events populated', async () => {
+      const meatbag = await Seeder.seedUser()
+      const device = await Seeder.seedUserDevice(meatbag, { deviceTraits })
 
-    const fetchedMeatbag = await getWithEvents(meatbag.id, usersDB)
```
vdiezel: We should consequently replace all references to `user` by `meatbag`.

lib/functional-helpers/mixin.js
DanielSeehausen: yes good idea i will update

lib/models/Connectable.js
DanielSeehausen: this is much better good :eye:

lib/models/Event.js
DanielSeehausen: resolved to `schema`, with `Class.SCHEMA` being the getter for the schema

lib/functional-helpers/mixin.js
DanielSeehausen: ill remove - looks like its been in node a long time now

PR #226, opened 2020-02-07
212 create event service with delete

### LINKED ISSUE #213: 7 alert: create DELETE /alert/<id>
in a new service

DanielSeehausen:
closes #213


lib/models/Event.js
```diff
@@ -4,6 +4,9 @@ const Joi = require('@hapi/joi')
 const { createSchemaAssertion } = require('validation-lib')
 const Connectable = require('models/Connectable')
 const mixin = require('functional-helpers/mixin')
+const { PROCESS_ACTIONS, DEVICE_ACTIONS, createDefaultGroups, RESOURCE } = require('access-rights-lib');
```
vdiezel: None of imports in line 7 is used.

DanielSeehausen: good :eye: I forgot about that
removed!

PR #227, opened 2020-02-10
222 implement first set of logs fired

### LINKED ISSUE #222: Implement a first set of device logs created on the backend
when following events are happening: 

- process start
- process paused
- process resumed
- process finished

vdiezel:
closes #222 
can be reviewed commit wise.


services/device/serverless.yml
```diff
@@ -718,6 +692,8 @@ functions:
       DB_USER: ${{file(../../env.${{self:provider.stage}}.yml):DB_USER}}
       DB_PASSWORD: ${{file(../../env.${{self:provider.stage}}.yml):DB_PASSWORD}}
       DB_NAME: ${{file(../../env.${{self:provider.stage}}.yml):DB_NAME}}
+      STAGE: ${{self:provider.stage}}
+      OSPIN_AWS_REGION: ${{file(../../env.${{self:provider.stage}}.yml):OSPIN_AWS_REGION}}
```
vdiezel: SNS needs them for message publishing

services/device/src/logger/logger.js
```diff
@@ -0,0 +1,50 @@
...
+
+    const process = await this.processesDB.findOne({ id: processId })
+
+    const msg = {
+      deviceId: device.id,
+      processId,
+      level: 'information',
+      subject: 'device/operation',
+      message: `Process ${process.name} ${transition}`,
+      createdBy: device.id,
```
O-Susanoo: i think createdBy should be the username right ?

vdiezel: No:  strictly the device is the source of truth and we don't actually know here which user started, finished, paused or resumed the process. The user logs should be created when a user tries to do that when he fires the commands and they could be something like `User X requested a start of process Y on device Z.

DanielSeehausen:
> Awesome so now we will start sending sms to users , that's cool

@O-Susanoo perhaps sometime soon in the future, that would be cool!
for now, this is for communication between lambda services

https://docs.aws.amazon.com/sns/latest/dg/sns-lambda-as-subscriber.html

services/device/src/logger/logger.js
```diff
@@ -0,0 +1,50 @@
...
+      case (device.state === 'running' && newState === 'idle'):
+        return 'finished'
+      case (device.state === 'running' && newState === 'paused'):
+        return 'paused'
+      case (device.state === 'idle' && newState === 'running'):
+        return 'started'
+      default:
+        return null
+    }
+  }
```
DanielSeehausen: this may be something the `Device` model should own

services/device/src/logger/logger.js
```diff
@@ -0,0 +1,50 @@
...
+
+  async updateState({ device, newState, runningProcessId }) {
+
+    const transition = Logger.getTransition(device, newState)
+
+    if (!transition) return
+
+    const processId = transition === 'finished'
+      ? device.runningProcessId
+      : runningProcessId
```
DanielSeehausen: this jumps out, but I don't doubt its here for a reason. 

seeing the `runningProcessId` passed, what is the corner case that is requiring the logging function to check this?

Is there a way to get the logger to do less work a device should perhaps own, and ensure whatever runningProcessId it gets is correct?

services/device/src/logger/logger.js
```diff
@@ -0,0 +1,50 @@
+const SNSLogAPI = require('SNSLogAPI')
+
+class Logger {
+
+  constructor(db) {
+    this.processesDB = db.collection('processes')
+  }
```
DanielSeehausen: what would you say to using `connectable` and mixin to get this a little closer to functional?

services/device/src/logger/logger.js
DanielSeehausen: makes sense
to build on this, a device can transition states without explicit user interaction (change phase, end process, etc.)

services/device/src/logger/logger.js
```diff
@@ -0,0 +1,50 @@
...
+    const msg = {
+      deviceId: device.id,
+      processId,
+      level: 'information',
+      subject: 'device/operation',
+      message: `Process ${process.name} ${transition}`,
+      createdBy: device.id,
+    }
+
+    return SNSLogAPI.publishLog(msg)
```
DanielSeehausen: I like the brevity, but chaining returns can sometimes obscure expected values (demeter and all)

It also means every thing that calls `updateState` needs to change if the return value of `SNSLogAPI` changes, as opposed to only changing how this method handles the return value

is there anything we can destructure out of this with an await and then return that explicitly?

services/device/src/logger/logger.js
```diff
@@ -0,0 +1,50 @@
...
+      case (device.state === 'running' && newState === 'paused'):
+        return 'paused'
+      case (device.state === 'idle' && newState === 'running'):
+        return 'started'
+      default:
+        return null
+    }
+  }
+
+  async updateState({ device, newState, runningProcessId }) {
```
DanielSeehausen: the name belies that `Logger` is smarter than it perhaps should be re: a device

services/device/src/logger/logger.js
```diff
@@ -0,0 +1,50 @@
+const SNSLogAPI = require('SNSLogAPI')
+
+class Logger {
```
DanielSeehausen: I believe the file name `logger.js` should follow casing of the class: `Logger.js`

services/device/src/device/triggers/updateState.js
```diff
@@ -24,6 +25,10 @@ async function updateState(event) {
   const device = new Device(db)
 
   const result = await device.updateState(requestData)
+
+  const deviceLogger = new Logger(db)
```
DanielSeehausen: <3 the use of logger in the controller
would also like to put my horse in the race for `connectable`

services/device/src/logger/logger.js
DanielSeehausen: if there is any chance this will be used from other services, consider placing it in `Models`. If our logger grows, something we may consider doing is having a `Logger` model, with inheriting children `DeviceLogger`, `ProcessLogger`, etc.

vdiezel: I'd like to follow the appraoch and not abstract this yet because it's to early to tell what pattern would be the best. I agree on the naming though.

PR #232, opened 2020-02-11
208 seperate assertions from protocol

### LINKED ISSUE #208: Differentiate MQTT lambda error handlers from HTTP error handlers and refactor how validation-lib creates errors
Current state:
- validations are run
- assertions (aka functions `assert/assertMany` uses) have internal knowledge of a handling function
- these handling functions are HTTP specific

Problem:
In the process of executing a lambda function that was fired from MQTT, errors may throw as a result of validators, and these errors are HTTP specific. This isn't appropriate with MQTT

Working solution:
assert/assertMany receives information about the protocol it is being used with. This is a problem because the models now know about their protocol, and we can't re-use functions between protocols :(

General Solution:
If errors throw from validators, they should be general. At the middleware level (which IS cognizant of its protocol, and should be) they are to be caught, logged, digested, and then either allowed to throw or be handled based on the protocol. 

This may seem like a lot, but with a little bit of planning it should be a pretty straightforward refactor. There will potentially be many files changed, but that is due to middlewares being added/removed from existing controllers.

vdiezel:
closes #208 

review commit wise. Assertions now throw a new Error called `HandledError` which is then caught in a new middleware called `mapErrorToProtocol` that takes a protocol (defaults to http). It maps the error into a shape which then can be used by a protocol specific middleware called afterwards, like `httpErrorHandler`. I've created two other handlers, one for `mqtt` and one for `sns` but they currently just log the error to the console. 
In the future one could merge all these middleware into a single big one that handles the errors, but for now it's good to go.

### update 
implemente middleware packages for HTTP, MQTT, and SNS
Start from here: https://github.com/ospin-web-dev/hambda/pull/232/commits/123484ba80817cf0a3145947fbf3fe6dbab16ad2

Bildschirmfoto von 2020-02-20 15-23-49


lib/error-lib.js
```diff
@@ -0,0 +1,26 @@
+class ErrorWithErrorCode extends Error {
```
vdiezel: is renamed to `HandledError` in the later commits and uses an error map instead of the number 1-5 for the error codes.
This error is thrown when assertions fail.

lib/validation-lib-v2.js
```diff
@@ -0,0 +1,217 @@
...
+ * }
+ * 'assert' functions make use of these response objects
+ *
+ * assert[Many] throws an http error (which middleware should handle)
+ */
+
+function assert(validator) {
+  if (typeof validator !== 'function') throw new Error('validator has to be a function')
+
+  const { invalid, errorMsg, errorFn } = validator()
```
vdiezel: the error function throws the new Error class with the correct error code.

lib/validation-lib-v2.js
```diff
@@ -0,0 +1,217 @@
...
+}
+
+exports.isResourceAdmin = isResourceAdmin
+
+exports.userHasProcessAccess = (device, process, userId, action) => (
+  or([
+    () => hasSufficientResourceAccess(process, userId, RESOURCE.PROCESS, action),
+    () => isResourceAdmin(device, userId),
+  ])()
+)
```
vdiezel: filename will go to `validation-lib` after refactoring

lib/middleware/mapErrorToProtocol.js
```diff
@@ -0,0 +1,18 @@
+const { createHTTPErrorResponse } = require('response-lib')
+
+module.exports = (protocol = 'http') => ({
+  onError: (handler, next) => {
+
+    if (handler.error.code && handler.error.message) {
```
vdiezel: handler.error.code will be true if it an error of the new created Error class, so all errors with code on them are considered handled!

services/device/test/trigger/createEvent.test.js
```diff
@@ -47,7 +47,9 @@ describe('device createEvent trigger', () => {
       describe('with a valid payload', () => {
         test('does not create the event', async () => {
           const payload = getValidPayload()
-          await fireCreateEventEvent(payload)
+          try {
+            await fireCreateEventEvent(payload)
+          } catch (_) { }
```
vdiezel: assertion of the correct error message could be added to the catch body in the future!

lib/middleware/snsErrorHandler.js
```diff
@@ -0,0 +1,10 @@
+module.exports = ({ logger = console.error } = {}) => ({
```
vdiezel: same as mqqtErrorHandler, just logs the error if it occurs

test/lib/validator.test.js
```diff
@@ -136,3 +136,121 @@ describe('assert',  () => {
...
+    test('should throw an Error', () => {
+
+      const resourceId = uuidv4()
+      const resource = undefined
+
+      expect(() => {
+        assertMany(() => canFind(resource, `Resource ${resourceId}`))
+      }).toThrow()
+    })
+  })
```
DanielSeehausen: nice

test/lib/validator.test.js
```diff
@@ -136,3 +136,121 @@ describe('assert',  () => {
...
+
+    test('should return empty array for a passed validation', () => {
+
+      const resourceId = uuidv4()
+      const resource = { id: resourceId }
+
+      const errorMessages = assertMany([ () => canFind(resource, `Resource ${resourceId}`) ])
+      const expected = []
+
+      expect(errorMessages).toEqual(expect.arrayContaining(expected))
```
DanielSeehausen: as im reading it here, it looks like `assertMany` doesn't return anything
[see here](https://github.com/ospin-web-dev/hambda/pull/232/commits/7c36f008c1e9b9fff3a8ceb82e263b5886e5c592#diff-e73972bf43a72ab4badf7b9216273694R33)

I don't understand how its passing as is though

should this expect instead be an expect to not throw error when run?

test/lib/validator.test.js
```diff
@@ -136,3 +136,121 @@ describe('assert',  () => {
...
+    })
+  })
+
+  describe('when used with multiple validation functions', () => {
+    test('should return a correct response when both validation fail', () => {
```
DanielSeehausen: I believe assertMany will fail fast on the first failed validation
(to collect all failures was the role of the deprecated `validateMany`, iirc)

test/lib/validator.test.js
```diff
@@ -136,3 +136,121 @@ describe('assert',  () => {
...
+  describe('when not providing the a functions as an argument', () => {
+    test('should throw an Error', () => {
+      expect(() => {
+        assert(3)
+      }).toThrow()
+    })
+  })
+
+  describe('when a valid validation function is provided', () => {
+    test('should return an error message when it fails', () => {
```
DanielSeehausen: believe this should throw, as opposed to return

test/lib/validator.test.js
```diff
@@ -136,3 +136,121 @@ describe('assert',  () => {
...
+
+      try {
+        assert(() => canFind(resource, `Resource ${resourceId}`))
+      } catch (e) {
+        const expected = `Can\'t find: Resource ${resourceId}`
+        expect(e.message).toEqual(expected)
+      }
+    })
+
+    test('should return undefined when it succeeds', () => {
```
DanielSeehausen: I believe the most important behavior we want to assert is that it does not throw error (checking undefined is an indirect way of testing that same thing)

test/lib/validator.test.js
```diff
@@ -136,3 +136,121 @@ describe('assert',  () => {
...
+    test('should return an error message when it fails', () => {
+      const resourceId = uuidv4()
+      const resource = undefined
+
+      try {
+        assert(() => canFind(resource, `Resource ${resourceId}`))
+      } catch (e) {
+        const expected = `Can\'t find: Resource ${resourceId}`
+        expect(e.message).toEqual(expected)
+      }
```
DanielSeehausen: with all of these try catch, jest has a way of expressing this behavior:
https://jestjs.io/docs/en/expect#tothrowerror

lib/response-lib.js
```diff
@@ -1,47 +1,6 @@
...
-}
-
-exports.conflict = (error) => {
-  const message = `Conflict: ${error}.`
-  throw new createError(409, message)
-}
-exports.iotMethodNotAllowed = (reason) => {
-  const message = `IoT Method Not Allowed Exception: ${reason}.`
-  throw new createError(405, message)
+exports.createHTTPErrorResponse = (protocol, error) => {
```
DanielSeehausen: I believe `protocol` is unused (and redundant here, because we are already in the create*HTTP*ErrorResponse

lib/middleware/mapErrorToProtocol.js
```diff
@@ -0,0 +1,18 @@
...
+      if (protocol === 'http') {
+        handler.error = createHTTPErrorResponse(protocol, handler.error)
```
DanielSeehausen: I believe there isn't a need to pass protocol here!

lib/middleware/mapErrorToProtocol.js
DanielSeehausen: I like the pattern a lot!

lib/middleware/logUnhandledErrors.js
```diff
@@ -8,9 +8,10 @@ module.exports = (opts) => {
 
   return ({
     onError: (handler, next) => {
-      const isHttpError = handler.error.statusCode
 
-      if (!isHttpError && typeof options.logger === 'function') {
+      const isHandledError = handler.error.code
```
DanielSeehausen: i like the pattern a bunch. Since you went through the trouble of creating the nice `HandledError` class, would you be opposed to adding something like a `handled` key ( = true) onto the object?

then the check becomes explicit, instead of relying on a side effect that when a code is present, it means it's handled.

lib/middleware/mapErrorToProtocol.js
vdiezel: :eye: :gun:

lib/middleware/mapErrorToProtocol.js
```diff
@@ -1,18 +1,26 @@
...
+          logger(handler.error)
+        }
       }
-    }
 
+      if (protocol === 'sns') {
+        if (typeof logger === 'function') {
+          logger(handler.error)
+        }
+      }
```
DanielSeehausen: would these log twice when we return `next(handler.error)` (once here, once when they hit `logUnhandledErrors` middleware if its present?). 

Reason I ask is because I see the http handler calls `next()`, and the other one looks like it passes the error on. I am not 100% on the middy API though, so let me know if I am mistaken

lib/middleware/stringifyResponseBody.js
```diff
@@ -12,10 +12,15 @@ const middleware = (handler, next) => {
...
 
 const errorMiddleware = (handler, next) => {
+
   if (!handler.response) {
     handler.response = {}
   }
 
+  if (handler.error.statusCode) {
+    handler.response.statusCode = handler.error.statusCode
+  }
```
DanielSeehausen: nice

DanielSeehausen: this looks a little out of place in `stringifyResponseBody`

services/device/src/command/endProcess.js
```diff
@@ -36,8 +35,7 @@ async function endProcess(event) {
 
 exports.main = middy(endProcess)
   .use(logUnhandledErrors())
-  .use(mapErrorToProtocol())
-  .use(httpErrorHandler({ logger }))
```
DanielSeehausen: I assumed you were going in the other direction, by keeping the `httpErrorHandler` and making duplicates for the other protocols, but I see why you went in this direction.

DanielSeehausen:
i like how code climate is complaining about something you brought up long ago @vdiezel re: combining our middlewares...

another day another time!

lib/middleware/mapErrorToProtocol.js
vdiezel: `logUnhandledErrors` is always called, but as the first middleware. Because the logging there only logs `unhandled` errors, while `mapErrorToProtocol` only does so for handled, it is not logged twice. 

Re: `next()` I'm confused myself. I replicated the pattern from `httpErrorHandler`. I tried calling `next(handler.error)` instead and It didn't make any diffrence.

services/device/src/command/endProcess.js
vdiezel: there was no way to keep `httpErrorHandler` without having an additonal middleware that actually creates the HTTP error. I could clone the existing one that would do that that and then have three different middleware again.

lib/middleware/mapHandledErrorToProtocol.js
```diff
@@ -0,0 +1,26 @@
+const { createHTTPErrorResponse } = require('response-lib')
+
+module.exports = ({ protocol, logger = console.error } = {}) => ({
```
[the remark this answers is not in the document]
vdiezel: :middle_finger:

O-Susanoo:
Viktor 81 files !!!!!!!!!!!!!!!!!!!!

lib/middleware/mapHandledErrorToProtocol.js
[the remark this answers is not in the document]
O-Susanoo: i see that codeclimate is your new best friend

lib/http-responses.js
```diff
@@ -0,0 +1,54 @@
...
+  const message = `No Authorization for this operation: ${reason}.`
+  return HTTPError(403, message)
+}
+
+const failedValidation = (reason) => {
+  const message = `Validation failed: ${reason}.`
+  return HTTPError(422, message)
+}
+
+exports.iotUnauthorized = (reason) => {
```
O-Susanoo: is there a reason why this is the only exported here ?

lib/middleware/HTTPDefaultMiddleware.js
```diff
@@ -0,0 +1,55 @@
...
+    useJSONBodyParser: true,
+    useSetRequestingUserId: true,
+    useHTTPEventNormalizer: true,
+    useStringifyResponseBody: true,
+    useCORS: true,
+  }
+
+  const options = Object.assign(defaultOptions, customOptions)
+
+  return Object.keys(middlewareMap).reduce((wrappedHandler, middleware) => {
```
O-Susanoo: i guess this one could be a separate function maybe ? build middleware ?

lib/http-responses.js
vdiezel: thats a good question. I think I didn't know where this was used and therefore didn't touch it yet because I was afraid of breaking something. I'll dive into it.

PR #234, opened 2020-02-11
Format client error stack trace

### LINKED ISSUE #233: format client error stacktrace in hambda

O-Susanoo:
closes #233


services/utils/src/logger/logger.js
```diff
@@ -4,13 +4,24 @@ const errorToSlackString = require('./serializers/errorToSlackString')
 
 class Logger {
```
DanielSeehausen: this snuck by first time, but just as a heads up - if a file is exporting a capital Class, it should be named the same 
`logger` --> `Logger`

services/utils/test/logger/logClientError.test.js
```diff
@@ -78,7 +85,7 @@ describe('Logger Class', () => {
       )
       expect(fetch).toHaveBeenCalledWith(
         expect.anything(),
-        expect.objectContaining({ body: expect.stringContaining(errorObject.stackTrace) })
+        expect.objectContaining({ body: expect.stringContaining(errorObject.fileName) })
```
DanielSeehausen: since we are testing for one of the new values, we should probably be checking for all of the new values

there may be a lot of duplication in these expects here!

PR #235, opened 2020-02-12
231 events close one another

### LINKED ISSUE #231: Certain events should resolve other events
Easiest described through example:

- device emits `doorOpen` event (I had an image linked from a certain 'artsy short film' here, of a man yelling, but I decided it better to leave it out.)
- event document exists in the Event collection, the alarm exists on the UI
- device emits a `doorClosed` event
- the event document is destroyed, the UI's alarm is resolved.

DanielSeehausen:
closes #231 

to keep the `Event` model true, made an `EventResolver`class, who owns all of the responsibility surrounding closing events.

where to actually act on revoking events was a difficult decision in terms of separation of concern for me. Ideally, separate endpoints would handle this, but because we want to reduce friction for event creators (think firmware team/the device itself), we opted to make the server smart. 

Now, the `device createEvent` endpoint is actually first checking whether the event to be created is a kind of event that resolves other events (think 'doorClosed' to an existing 'doorOpen'). If it is, we fork at that point, and attempt to revoke/close those matching existing events. If it isn't we make a new event, per the normal functioning of `createEvent`

this can be reviewed commit wise. Last commit which adds two tests is tough to look at re: the diff because of indentation stuff. Easiest to just look at the file plainly and view the newly added tests. 

this should be the closing ticket for our `Events` responsibilities for now (until we get to the min/max breaching stuff).


lib/models/Event/Event.js
```diff
@@ -31,7 +31,6 @@ class Event {
 
     if (eventTypeSupported) return eventType
 
-    console.warn(`${eventType} is not a supported event type!`)
```
DanielSeehausen: oops I did not mean to remove this

services/device/src/device/triggers/createEvent.js
```diff
@@ -16,6 +17,16 @@ const validateCreateEvent = require('../../joi/triggers').createEvent
 
 const Device = require('../device')
 
+async function revokeEvents({ revokingName, deviceId }) {
+  const revokedEventIds = await EventRevoker.revokeMatchingEvents({ revokingName, deviceId })
+
+  revokedEventIds.forEach(revokedId => {
+    IotDataAPI.publishDeviceEvent(deviceId, { revokedId })
```
vdiezel: will this work without await?

services/device/src/device/triggers/createEvent.js
```diff
@@ -31,15 +42,18 @@ async function createEvent(requestData) {
   const db = await mongo.getDB()
   const device = new Device(db)
   Event.connect(db)
+  EventRevoker.connect(db)
+
+  const eventRevokesEvents = EventRevoker.isRevokingEventName(eventData.name)
+  if (eventRevokesEvents) {
+    return revokeEvents({ revokingName: eventData.name, deviceId: eventData.deviceId })
```
vdiezel: don't you have to await this here?

DanielSeehausen: discussed in person

services/device/src/device/triggers/createEvent.js
DanielSeehausen: discussed in person - fire and forget!

lib/models/Event/EventRevoker.js
```diff
@@ -0,0 +1,51 @@
...
+  static get REVOCATION_MAP() {
+    return {
+      doorClosed: new Set(['doorOpen', 'doorOpened']),
+      doorClose: new Set(['doorOpen', 'doorOpened']),
+    }
+  }
+
+  static isRevokingEventName(eventName) {
+    const revokingNames = new Set(Object.keys(EventRevoker.REVOCATION_MAP))
+    return revokingNames.has(eventName)
```
vdiezel: `return eventName in EventRevoker`

DanielSeehausen: very nice - updated

lib/models/Event/EventRevoker.js
```diff
@@ -0,0 +1,50 @@
...
+
+  static getRevokedEventNames(revokingName) {
+    return EventRevoker.REVOCATION_MAP[revokingName] || []
+  }
+
+  static getRevocableEvents({ revokingName, deviceId }) {
+    const revocableNames = EventRevoker.getRevokedEventNames(revokingName)
+
+    return EventRevoker.eventsDB.find({
+      name: { $in: Array.from(revocableNames) },
```
O-Susanoo: if mongo goes to run this query even if revocableNames is empty array than i think we should only call this find if revocableNames is not empty

DanielSeehausen: good catch @O-Susanoo 
updated!

DanielSeehausen:
this is ready to go now - please see last commit!

DanielSeehausen:
this is ready to go now! had a bug and was unsure if the lambda was killing early - so I refactored and ended up awaiting everything. Everything seems to be working as intended.

PR #253, opened 2020-02-24
Implements a backend service CORTEX for parsing datapoints and firing events

### LINKED ISSUE #247: Creata a backend service to parse incoming datapoint and create alerts when triggers hit

vdiezel:
closes #247


services/cortex/models/EventTriggerManager.js
```diff
@@ -0,0 +1,112 @@
...
+    if (!registeredTriggers.length) return []
+
+    const firingTriggers = EventTriggerManager
+      .getFiringTriggers(registeredTriggers, parsedMsg, process)
+
+    EventTriggerManager.updateEventTriggerTimestamps(firingTriggers, process)
+
+    return EventTriggerManager.mapTriggersToEvents(firingTriggers, process)
+  }
+}
```
vdiezel: logic is this: 
1. Get all triggers that can be found for the currently running phase in the process description
2. 2 Step evaluation if they should create an event: 
    2.1 Check if the trigger condition is true (e.g. is the value above or below the user set limit)
    2.2 Check if the trigger has already fired, if so, check if it should fire again based on the `DEBOUNCE_TIME`

services/cortex/models/MQTTEventHandler.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  static parseMessageToObject(msgBuffer) {
+    return JSON.parse(msgBuffer.toString())
+  }
+
+  publishEvents(events) {
+    events.forEach(event => {
+      console.log('publishing', event)
+      //this.MQTTClient.publish(`${process.env.STAGE}/device/${event.deviceId}/event/create`, JSON.stringify(event))
```
vdiezel: commented out for testing purposes

services/cortex/test/EventTriggerManager.test.js
```diff
@@ -0,0 +1,396 @@
...
+
+      const addPhaseMock = jest.fn()
+      ProcessUtils.addPhase = addPhaseMock
+
+      EVM.updateProcessPhase(mqttMessage)
+
+      expect(addPhaseMock).not.toHaveBeenCalled()
+    })
+  })
+})
```
vdiezel: turned out pretty verbose because the whole trigger deal needs a lot of setup

services/cortex/models/Event/LowerLimitTrigger.js
```diff
@@ -0,0 +1,17 @@
...
+const EventTrigger = require('./EventTrigger')
+
+
+class LowerLimitTrigger extends OutOfBoundaryTrigger {
+
+  evaluate(values) {
+    const minValue = Math.min(...values)
+
+    if (minValue >= this.limit) return EventTrigger.nonFiringTrigger()
+    this.measuredValue = minValue
```
O-Susanoo: evaluate seem a little bit misleading because you are creating a side effect here , so according to the squirrel book aka functional thinking , either a more revealing name, or another pattern

services/cortex/models/EventTriggerManager.js
```diff
@@ -0,0 +1,112 @@
...
+
+  static getNotRecentlyFiredTriggers(triggers, process) {
+    return triggers.filter(trigger => {
+
+      const { type, streamId } = trigger
+      const firedEvent = ProcessUtils.getEvent(process, streamId, type)
+
+      if (!firedEvent) return true
+
+      if (Date.now() - firedEvent.firedAt > EventTriggerManager.DEBOUNCE_TIME) {
```
O-Susanoo: Again as i mentioned in slack the issue with such behavior is that the event will keep firing the same thing even tough nothing has changed and this is due to the fact that we are checking the time not the state change

services/cortex/models/ProcessCache.js
```diff
@@ -0,0 +1,30 @@
+const LRU = require('lru')
+const ProcessUtils = require('./ProcessUtils')
+
+class ProcessCache {
+
+  constructor(size = 100) {
+    this.cache = new LRU(size)
```
O-Susanoo: cool love this stuff but never work on it , i'm taking some side reading to discover this awesome package

services/cortex/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "cortex",
+  "version": "1.0.0",
+  "description": "",
+  "main": "app.js",
+  "dependencies": {
+    "aws-mqtt": "^1.0.0",
+    "aws-sdk": "^2.619.0",
+    "lru": "^3.1.0",
```
DanielSeehausen: seeing this as a package name is getting me HYPED for this implementation

services/cortex/models/ProcessUtils.js
```diff
@@ -0,0 +1,62 @@
+class ProcessUtils {
```
DanielSeehausen: This file is showcasing the need to move `process/process` up to a top level process model. :O

While refactoring the existing process to be a model would be a lot of work, I'd propose just moving this up to `models/process/Process` for now, and we can refactor the other stuff into this later.

services/cortex/models/ProcessCache.js
```diff
@@ -0,0 +1,30 @@
...
+
+  contains(processId) {
+    return this.cache.peek(processId) !== undefined
+  }
+
+  retrieve(processId) {
+    return this.cache.get(processId)
+  }
+
+  add(processId, processData) {
```
DanielSeehausen: `add` may be misleading -
they use `set` on cache I imagine to imply if an object already exists on the identifier (the processId), that it is replaced. `add` makes it seem like the cache is accruing these values, instead of replacing them, or should at least fail or refuse if the thing already exists

services/cortex/app.js
```diff
@@ -0,0 +1,10 @@
+require('./startup/aws')
+const EventTriggerManager = require('./models/MQTTEventHandler')
+const MQTTEventHandler = require('./models/MQTTEventHandler')
```
DanielSeehausen: are these both supposed to require the same thing? I don't see the `models/EventTriggerManager` being used anywhere

services/cortex/models/MQTTEventHandler.js
```diff
@@ -0,0 +1,65 @@
+const getMQTTClient = require('../startup/MQTTClient')
+
+
+class MQTTEventHandler {
+
+  constructor(eventTriggerManager) {
+    this.MQTTClient = getMQTTClient()
+    this.EVM = eventTriggerManager
```
DanielSeehausen: Was this meant to be `ETM`?

services/cortex/models/EventTriggerManager.js
```diff
@@ -0,0 +1,112 @@
...
+    return 1800000 // ms = 30 mins
+  }
+
+  updateProcessPhase(parsedMsg) {
+    const { processId, phaseId, startTime } = parsedMsg
+    const newPhase = { phaseId, startTime }
+    return this.processCache.updatePhase(processId, newPhase)
+  }
+
+  static getNotRecentlyFiredTriggers(triggers, process) {
```
DanielSeehausen: organization may benefit from grouping static and instance methods (traditionally, static at the top, but that doesn't really matter)

services/cortex/models/ProcessUtils.js
```diff
@@ -0,0 +1,62 @@
...
+    const currentPhase = this.getCurrentPhase(process)
+    if (!currentPhase) return []
+
+    const phaseDownStreams = this.getPhaseDownStreams(process, currentPhase.phaseId)
+    const downStreamIds = Object.keys(phaseDownStreams)
+
+    return downStreamIds.reduce((acc, streamId) => {
+      const streamEventTriggers = phaseDownStreams[streamId].eventTriggers
+
+      if (streamEventTriggers && streamEventTriggers.length) {
```
DanielSeehausen: can the shorthand operator be used here for this? I don't remember what its called, but looks something like:
`streamEventTriggers&.length`

services/cortex/models/MQTTEventHandler.js
```diff
@@ -0,0 +1,65 @@
...
+          this.EVM.updateProcessPhase(parsedMsg)
+          break
+        }
+        case 'dataPoints': {
+          const triggeredEvents = this.EVM.getTriggeredEvents(parsedMsg)
+          if (!triggeredEvents || !triggeredEvents.length) return
+
+          this.publishEvents(triggeredEvents)
+          break
+        }
```
DanielSeehausen: breaking these out into functions will clear up a big potential confusion, because it provides a cheap change to use different names for `parsedMsg`

services/cortex/models/DataUtils.js
```diff
@@ -0,0 +1,14 @@
+class DataUtils {
+
+  static parseStreamData(parsedMsg) {
+    // we just get rid of processId, messageSequence and lastMessage
+    const { processId, messageSequence, lastMessage, ...streamData } = parsedMsg
```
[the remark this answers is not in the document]
vdiezel: no, I use destructuring to actually get rid of the keys I don't need, leaving me with the stream data in form of 
```javascript 
{
'1213132': [ [ times ], [ values ] ],
'1381228': [ [ times ], [ values ] ],
....
}
```

services/cortex/models/ProcessUtils.js
vdiezel: NodeJS does not support optional pathing like `myArray?.length` yet.

services/cortex/models/EventTriggerManager.js
vdiezel: I can't follow you here. The event will be only fired agan if it was not fired for at least 30 minutes.

vdiezel:
> `.gitignore` should perhaps be copied from a service (or made) to ignore `package-lock` and `node_modules`
> 
> (while git can reference up to the top one, when working in the service directly it is not used)
> 
> On a similar note, do we want a default `npm start` script here?

done!

services/cortex/models/MQTTEventHandler.js
vdiezel: has already been done!

services/cortex/models/MQTTEventHandler.js
vdiezel: EVM ist just short for eventTriggerManager - I could take the longer name back in.

services/cortex/models/EventTriggerManager.js
vdiezel: And we are not going to resolve the event automatically - it will be done manually by the customer

services/cortex/models/ProcessCache.js
vdiezel: you are right

services/cortex/models/Event/LowerLimitTrigger.js
vdiezel: Yes, I was aware of this - guess I gotta try harder.

vdiezel: fixed.

services/cortex/models/ProcessUtils.js
vdiezel: Done. I moved the general part to `models/Process/Process`; The logic that was acting on the triggeredEvents that I was adding to the process model at runtime in the cache was abstracted to the `EventTriggerManager` - its suited better there and this way we are not messing up the universality of the `Process` interface.

services/cortex/package.json
```diff
@@ -17,8 +17,7 @@
     "uuid": "^3.4.0"
   },
   "scripts": {
-    "seed-lib-files": "cp -rf ../../lib/* ./node_modules/",
-    "seed-lib-files-temp-fix": "cp -rf ../../lib/* ./shared/",
+    "seed-lib-files": "cp -rf ../../lib/* ./lib/",
```
DanielSeehausen: v nice - would you be able to make a devops ticket to update this everywhere else as well? I believe this is a much better structure to get in to and less error prone!

DanielSeehausen: this comment might have gotten moved somehow - I mean making a ticket re: the cp of lib files instead of seeing?

vdiezel: I'll do it

PR #254, opened 2020-02-25
236 events create logs

### LINKED ISSUE #236: All events should create a log
As a user, whenever a device event is created, of any kind (alarm, notification, warning), I should see this event logged in the device logs.

- [x] all event creations should create logs via calling the log service
- [x] event revocations should create a log
- [x] since events revoke on device start up, a log should be provided to clearly state that the device has been restarted (and possible also that all events are revoked)
- [x] make sure it shows up on the front end!

DanielSeehausen:
Closes #236 (has mutliple checkboxes)

the commits are in decent enough shape - much easier to go through there than files changes

Here is what the logs look like:

image


services/device/src/device/triggers/createEvent.js
```diff
@@ -9,9 +9,18 @@ const Device = require('../device')
...
+  const revokedEventIds = revokedEvents.map(event => event.id)
+
+  await IotDataAPI.publishDeviceEventRevocations(deviceId, revokedEventIds)
+
+  const revokedEventNames = Event.getUniqueNames(revokedEvents)
+  await Logger.logEventsRevoked({
+    deviceId,
+    processId,
+    message: `${revokingName} revokes events: ${revokedEventNames.join(', ')}`,
+  })
```
vdiezel: Do we really want to show the event revocations to the customer? I think it would be sufficient just to show him what events where fired. All a customer wants to know is when the door was closed. Do they really care that it revoked all previos `doorOpen` events? I don't think so - all they care about are the events expressed in a nice message with their timestamps.  would `resolving` alarms in the log not make more sense than act would `resolving` alarms in the log not make more sense than act would `resolving` alarms in the log not make more sense than act

I also think that we don't need to tell them that we revoked all events on startup - It might be interesting however to log when the device appeared online after a startup - I think this would deserve a log.

My point is to hide the idea of event revocations from the customer - all they care about are events. Event revocations is something merely in the internal working of our software.

Also, would it make more sense to talk about `resolved alarms/events` instead of `revoked events , at least in the UI?

What your opinion on this guys?

DanielSeehausen: Your focus on what matters to the user is strong - I agree with you re: the log having unnecessary implementation. I'd like to take it out as well. Let me provide my explanation why I put it in to be transparent and make sure the use case we lose is acceptable:

My thought was, if the user has logs of multiple alarms firing, they might want a record of when those alarms were resolved as well. Following that train of thought, I wanted to make sure no alarm got closed/resolved without their being a record of when it closed/resolved (i.e. all alarm resolutions have a log, so that process history can be accurately analyzed through the logs). 

This was my solution to ensuring that the record was preserved. Without it, events that resolve multiple alarms might not be obvious to a user. E.g. something like `doorClosed` also resolved some other sensor alarm that was active. From a user's perspective of reading the log as a history, they could easily assume `doorClosed` means the opposite of `doorOpen`, and therefore it's resolved at that point in history. They wouldn't necessarily be able to associate `doorClose` with the resolution of another alarm.

> My point is to hide the idea of event revocations from the customer - all they care about are events. Event revocations is something merely in the internal working of our software.

I agree re: hiding the revocation concept. I think we should still log revocations as events then, with just more user friendly messaging (for the history purposes described above)

DanielSeehausen:
> I love the clean implementation! Nothing blocking from my side.
Thank you. We are all short story authors, with two dedicated and passionate fans each (soon to be a gang of 4).

> I left a big comment on the desired behaviour for the logging and the kind of messages we want to show to the cusomter to keep noise out.
Responded on the comment!

> Also we really need something to map the events `doorOpen` etc. to more user friendly messages.
I am going to add this

DanielSeehausen:
updated with couple more commits on top. should be easy to read in order!

PR #258, opened 2020-02-28
fixes the reference to log js in create and list

vdiezel:


vdiezel:
This fix has already been deployed to dev and prod!

DanielSeehausen:
Would you be able to invert the fix here -- 

The file name should be `Log`, as it exports the capital class `Log` !

PR #272, opened 2020-03-04
270 cli deployment for elastic beanstalk

### LINKED ISSUE #270: Create a CLI deployment suite for cortex
Currently, Cortex is updated using the AWS interface and manually uploading a zip file there. This should be automated.

vdiezel:
Implements a CLI to deploy cortex service and dev and prod environments:
closes #270 

Easiest to review the final result.

## SETUP
Cortex is deployed on Elastic Beanstalk (EB) in region us-west-2, the only region EB supports:

https://us-west-2.console.aws.amazon.com/elasticbeanstalk/home?region=us-west-2#/application/overview?applicationName=cortex

There we have two environments `dev` and `prod` for an application called `cortex` -  they run a NodeJS environment for us. The code is required via S3 - our builds are stored in a bucket called `cortex-builds`.

## DEPLOYMENT

The code needs to be bundled in a zip/war file (`build` step). Then it is pushed to the deployment bucket and a new application version is created which takes the new bundle as its source code (`deploy` step). The last step is to update the desired environment to use the newly created application version (going to a previous version is also possible for a rollback). Only with the last step the running code on the instances will actually be updated. That is the `release` step. 

I followed the recommended pattern and used the last git commit hash as a version label. which means, whenever you want to deploy new code you also have to create a new commit, otherwise AWS will reject the new application version due to duplicate versions. Lets see if this is too strict and annoying during development - then we could just randomly generate version-labels with `uuid`.

I modifed our circleCI user to be able to run this pipeline, too. For deployment from your local machine your local `default` profile in `/home/.aws/credentials` should be a user with sufficient access. Most likely you will have some admin user set there which should be sufficient. If not, either get access key id and secret of a user that is powerful enough (either from AWS IAM or our Confluence crendetials page) or create a new user which has the right privileges. You need the same access as the IAM user `circleCI` has.

## MONITORING

EB provides a nice UI where you can see the progress after the release. It also gives you some application health hints. I activated logging to cloudwatch for both environments - just go to CloudWatch, switch region to `Oregon (us-west-2)` and check the logs.


services/cortex/package.json
```diff
@@ -14,13 +14,21 @@
...
     "postinstall": "npm run seed-lib-files",
     "pretest": "npm run seed-lib-files",
-    "test": "jest -i"
+    "test": "jest -i",
+    "build": "nps build",
+    "deploy": "nps deploy",
+    "releaseToDev": "nps releaseToDev",
+    "releaseToProd": "nps releaseToProd",
+    "releaseEverywhere": "nps releaseToDev releaseToProd",
+    "rolloutDev": "nps build deploy releaseToDev",
```
vdiezel: a rollout only on `prod` didn't make sense to me, because I didn't want to encourage skipping deploying the same code on dev.  I was struggling with the naming for running the full build-deploy-release pipeline. Called it `rollout`; also thought of `runCDSuite` or `ship`...

O-Susanoo: ship would have been way cooler , specially with the new terminal you can actually print the ship icon :+1:

DanielSeehausen:
strong memeing mo

have you had a chance to test the latencies with US-West @vdiezel? That is an intimidating distance, but if its not a problem its not a problem. This might be a good chance to integrate datadog. It's unclear to me whether you just need to enable some settings for reporting, or whether you need to actually make sure the client is present on the deployed server:

https://docs.datadoghq.com/integrations/amazon_elasticbeanstalk/

vdiezel:
> Looks great - have requested changes by they aren't re: issues with the code
> 
> there is a lot of information in this PR that shouldn't go poof once it gets merged.
> Would you be able to make a readme in the project directory that condenses what you wrote in the PR?
> 
> Unsure if this PR/issue is the right place, but would you be able to hook this up to datadog?
> 
> This CLI deployment looks great. Is it hooked up to our continuous delivery in circleCI? If not, would you be able to do that? We are taking on water already doing this with a team of our size with the other lambda services, and CD really pays dividends. using:
> `git show --name-only | grep services/cortex/` might be a janky way to recognize something has changed in cortex! That was my half baked plan for service specific deploys later
> 
> If you think either or both of the above requests belong in their own tickets - would you be able to create them? We can put them in the new sprint

Thanks for the review. I think the README should go into this PR, while the logging to datadog and the CD implementation of cortex should go into seperate tickets!

PR #273, opened 2020-03-04
264 make collections for users unique for emails usernames

### LINKED ISSUE #264: Make users collections unique index on name, id, and email
**Pre-requisites**
- Either use the mongoDB atlas interface or (recommended) download a mongo client (e.g. MongoDB Compass Community)
- get login credentials from Viktor

**Deliverables**
- [ ] unique index the `users` collection `userName`, `email`, and `id` (the should already have a unique index on the `_id`)
- [ ] update the create user method to return either success OR 409 if already exists (which will need to be determined from the mongodb insertOne response in create). In order to raise the 409 add a validator that raises a 409. Check if the existing http error handler automatically appends appropriate text to a 409. If not, provide it! (ping viktor or daniel for pairing on the validator pattern). 

Currently, the method in which we create users (via AWS Cognito) will reject creating duplicate users by both email and userName, but we want to ensure our DB does the same independently of the method of creation.

ConorTighe1995:
closes #264 

The Mongo database for the Ospin web application should only accept user documents that have unique values for both email and username. Indexes are now applied to insure there are no duplicated values in dev or prod. Below is an example of how I applied the index for unique values to id,email and username.

image

Changes were also made in the hambda server to handle Mongo related errors thrown. In the case of an attempt to write to a unique field where the value in the user object already exists in a document inside the user collection, MongoDB will return an error with the error code of 11000. This tells the hambda server that a the data already exists in the database and a duplication error is created. The created error is passed to error-lib by calling **createDuplicateError** where it be passed to the HandledError constructor and handled.


services/user/src/user/create.js
```diff
@@ -3,6 +3,8 @@ const mongo = require('mongodb-lib')
 const validateCreate = require('../joi/user').create
 const { success } = require('http-responses')
 const HTTPDefaultMiddleware = require('middleware/HTTPDefaultMiddleware')
+const { throwDuplicateError } = require('error-lib')
```
vdiezel: This is not used here!

services/user/test/user.test.js
```diff
@@ -80,12 +80,59 @@ describe('User Class', () => {
 
   describe('create',  () => {
 
+    beforeAll(async () => {
+      await usersDB.createIndex({ id: 1 }, { unique: true });
+      await usersDB.createIndex({ email: 1 }, { unique: true });
+      await usersDB.createIndex({ username: 1}, { unique: true });
+    })
```
vdiezel: :+1:

PR #283, opened 2020-03-06
275 implement models for cortex dp

### LINKED ISSUE #275: DATAPOINTS: STORAGE I Create Models for CacheManager and RunningProcessInterface
Datapoints that are received on cortex should be stored in a cache until enough datapoints are collected to create a datachunk that will be stored in the database.To begin with, the data should be stored locally in the node RAM for this ticket. 

This ticket includes: 
- [ ] writing tests to cover the desired behaviour
- [ ] Write up the models for the CacheManager und RunningProcessInterface 
- [ ] Connect the new models to the existing `MQTTEventManeger`

The logic for storing data in MongoDB should not be covered here yet. Write up a an empty function for this and simply assure in tests thta it is called. 

image

vdiezel:
closes #275 

Integrates first drafts for the storage side of the new cortex datapoints service:

- RunningProcessInterface takes care of incoming data, storing it in the cache and dumping the data into the database (not implemented) when we have collected enough data for a chunk
- CacheManager stores remaing data of the last data chunk when a process has finished and also takes care of removing processes from the cache that went stale. Processes could go stale when e.g. the internet connection breaks for more than a day or the device was shutdown not correctly. 

the cache service is just spoofed for now. 

Review commit wise is possible.


services/cortex/models/RunningProcessInterface.js
```diff
@@ -0,0 +1,94 @@
...
+      cachedData[streamId].x.push(dp)
+    })
+  }
+
+  static appendYValues(cachedData, streamId, data) {
+    data.forEach(dp => {
+      cachedData[streamId].y.push(dp)
+    })
+  }
+
```
vdiezel: didn't combine the old and the new data using spread operator because I was worried about perfromance here

services/cortex/models/RunningProcessInterface.js
```diff
@@ -0,0 +1,94 @@
...
+    const streamIds = Object.keys(data)
+
+    const cachedData = cache.get(processId) || {}
+
+    streamIds.forEach(streamId => {
+      this.handleStreamData(processId, cachedData, streamId, data)
+    })
+
+    this.updateTimestamp(cachedData)
+    cache.set(processId, cachedData)
```
vdiezel: the `cache.set` is redundant here because we update by reference, but with a cache service it will be necessary again, so I left it here as a memo.

services/cortex/models/CacheManager.js
```diff
@@ -0,0 +1,64 @@
...
+    return 24 * 60 * 60 * 1000 // 1 day
+  }
+
+  static getAllProcesses(cacheServiceSpoof) {
+    return Object.keys(cacheServiceSpoof.cache).map(processId => ({
+      processId, ...cacheServiceSpoof.cache[processId],
+    }))
+  }
+
+  static start(cache) {
```
O-Susanoo: i think what you are looking here for is setInterval , at least that's how i understood the goal of such manager it should after each 1h try to evacuate the cache or am i wrong

services/cortex/models/CacheManager.js
```diff
@@ -0,0 +1,64 @@
...
+    return setTimeout(
+      this.evacuateStaleData(cache),
+      this.INTERVAL,
+    )
+  }
```
O-Susanoo: i think what you are looking here for is setInterval , at least that's how i understood the goal of such manager it should after each 1h try to evacuate the cache or am i wrong

Issue #267, opened 2020-03-02
Copy files from the shared top level "lib" directory to local "lib" folders within the service


vdiezel:
duplicate #362 

We currently have a hacky solution for files that are used cross service. They are stored in the top level directory called `lib` and whenever we run tests or deploy we copy them into the `node-modules` folder of the service. 

This has a couple of disadvantages: When the files are stored in node modules they are imported like normal npm modules e.g. `const { createDefaultAccessGroup }  = require('access-rights-lib')`. This confuses eslint, because our own files are not listed as packages in package.json. 

Also, looking up the local files in your node_modules is annoying and `npm i` will blow them away.

Another reason is that AWS elastic beanstalk can't resolve single files from `node_modules`.

So, we want to move lib files now to `lib` folders within every service instead and reference them correctly within the service.

PR #288, opened 2020-03-10
276 implement caching service

### LINKED ISSUE #276: DATAPOINTS: STORAGE II Connect Cortex to Cache Service on AWS
AWS provides cache services. Part of this ticket is:

- [ ] Spike into possible solution for connecting Elastic Beanstalk to a cache service
- [ ] Implement the service into Cortex
- [ ] Modify CacheManager to use the new cache service

vdiezel:
closes #276 
 
We have now an AWS ElasticCache Redis running! 

I connected it to the cortex service and updated the code to use it instead of the CacheServiceSpoof! 

I also implemented something called 'Hooks' which are fired on process transitions , otherwise we would need to deduce the process transition by listening to device state updates which is pretty annyoing, we do it on the frontend for example - in future we could replace it there too by using the hooks. For now I just need the `process finished` hook to save remaining data from the cache to the database.


lib/models/UIConfig/UIConfig.js
```diff
@@ -1,10 +1,10 @@
...
-const DeviceDashboard = require('models/UIConfig/DeviceDashboard')
-const Graphs = require('models/UIConfig/Graphs')
-const Connectable = require('models/Connectable')
-const mixin = require('functional-helpers/mixin')
-const { assertMany, canFind } = require('validation-lib')
+const DeviceDashboard = require('../../models/UIConfig/DeviceDashboard')
+const Graphs = require('../../models/UIConfig/Graphs')
+const Connectable = require('../../models/Connectable')
+const mixin = require('../../functional-helpers/mixin')
+const { assertMany, canFind } = require('../../validation-lib')
```
vdiezel: all imports within libs should use relative paths because we are planning not to store our own lib files within node modules: https://github.com/ospin-web-dev/hambda/issues/267

services/cortex/models/CacheManager.js
```diff
@@ -11,53 +12,60 @@ class CacheManager {
...
-  static evacuateProcess(processData, cacheServiceSpoof) {
+  static async evacuateProcess(processData) {
     this.evacuateStreamData(processData)
-    this.removeFromCache(processData.processId, cacheServiceSpoof)
+    await this.removeFromCache(processData.processId)
   }
 
-  static evacuateStaleData(cacheServiceSpoof) {
-    const cachedProcesses = this.getAllProcesses(cacheServiceSpoof)
+  static async evacuateStaleData() {
```
vdiezel: Gets all processes from cache, checks if their haven't been updated for a certain amount, if so, save their data to the DB and delete the process from cache.

services/cortex/models/MQTTEventHandler.js
```diff
@@ -6,28 +10,64 @@ class MQTTEventHandler {
   }
 
   static parseMessageToObject(msgBuffer) {
-    return JSON.parse(msgBuffer.toString())
+    let msg
+    try {
+      msg = JSON.parse(msgBuffer.toString())
+    } catch (e) {
+      console.log(e)
+    }
```
vdiezel: guard for malformed mqtt messages to prevent app breaking

services/cortex/models/MQTTEventHandler.js
```diff
@@ -44,22 +84,33 @@ class MQTTEventHandler {
 
     this.mqttClient.on('connect', () => {
       console.log('Connected to MQTT-Broker.')
+
       this.mqttClient.subscribe(`${process.env.STAGE}/dataPoints/#`)
       this.mqttClient.subscribe(`${process.env.STAGE}/phases/#`)
+      this.mqttClient.subscribe(`${process.env.STAGE}/device/+/hooks/process`)
```
vdiezel: topic for process hooks is `{env}/device/{deviceId}/hooks/process`

services/cortex/startup/redis.js
```diff
@@ -0,0 +1,38 @@
...
+
+  const redisClient = redis.createClient({
+    host: HOST_URL,
+    port: PORT,
+  })
+
+  redisClient.set = promisify(redisClient.set)
+  redisClient.get = promisify(redisClient.get)
+  redisClient.keys = promisify(redisClient.keys)
+  redisClient.del = promisify(redisClient.del)
```
vdiezel: promisifying the redis calls to be able to use `async/await`

PR #293, opened 2020-03-16
291 use timestamp to sort datachunks

### LINKED ISSUE #291: Use the timeStamp property when sorting data chunks when fetching data from the DB
We are using the chunkNumber `n` to sort chunks when we fetch them from the database. The chunk number is needed for the the data storage implementation so we know which chunk to grow in the database when we insert data points.

With the new implementation the chunk number becomes abundant and will be removed. For sorting the chunks we simply can use the `t` property (the timestamp of the first value in the chunk).

vdiezel:
closes #291 
Review commit wise.


vdiezel:
> Good idea, what exactly are the data chunks of? are they just binaries of what is being stored in mongodb?
They are a collection of datapoints, stored in redis as stringified objects. 
The device publishes data every five seconds. To prevent going to MongoDB every 5 seconds we simply collect a certain amount of data points (currently 1800, which represents about 30 minutes of data for most sensors) in a so called 'chunk' and store this chunk finally in a single operation in MongoDB. 

the chunk size is a compromise: If we make them to big, we can't query data by time properly, if we make them too small, it increases the storage overhead. In the beginning we even thought of storing every datapoint seperately, but this was too expensive storage wise because every data point got a MongoID which grows the size of every data point to about 80-100 bytes.

Issue #299, opened 2020-03-19
fetching stream data for big data chunks causes a Mongo Error due to RAM restrictions 


vdiezel:
COMPLETION IS MAKING THE CHILDREN TICKETS THAT WOULD SOLVE THE PROBLEM AND DOCUMENTING THE PROCESS IN DETAIL

Certain streams do produce much more data than normal once, up to 200/second (!). When trying to fetch those in the `getDataPoints` function, Mongo immediately refuses to do so, giving this error:

![Bildschirmfoto von 2020-03-19 09-51-46](https://user-images.githubusercontent.com/37934155/77049129-d1f31a80-69c7-11ea-864f-71779bc981bf.png)

Investigate what makes it  fail and how to solve this.

O-Susanoo:
Closing this ticket , it was solved by adding this index processId_1_streamId_1_startTime_1 the issue was in this piece of code (dataPoints)

```
    const chunks = await this.dataPointsDB
      .find({ processId, streamId, startTime: { $lte: maxStartTime } })
      .sort({ startTime: 1 })
      .toArray()
```

to summarize because we didn't have a good index mongodb was forced to take out document to memory before sorting them but since there is a limit on memory usage it fails you can read more about it here

[Doc](https://developerslogblog.wordpress.com/2019/11/27/mongodb-sort-how-to-fix-maximum-ram-exceeded-error/)

if you are the type of guy who believe what they see then here is how you can create the issue and solve it by yourself

- run this command from a node script 
```
    const chunks = await this.dataPointsDB
      .find({ processId: '693e2ab3-0323-474f-86ec-b3b22b3ba8bb', startTime: { $lte: maxStartTime } })
      .sort({ startTime: 1 })
      .toArray()
```
693e2ab3-0323-474f-86ec-b3b22b3ba8bb is processId of a long running process (prod) 3d

you will get the error in this ticket now go to your mongo console and create an index that will have only processId and startTime and run the same command again then you will no longer seee the issue

(if you do so please don't forget to remove that index)

PR #300, opened 2020-03-19
fixes middleware error propagation

### LINKED ISSUE #297: When doanloading certain streams Lambda returns an Invoke error
An example is the `pressure up` stream in the heartbeat group of the FU bioreactor

Bildschirmfoto von 2020-03-18 08-12-46

vdiezel:
closes #297 

We had a couple of issues with our middleware: 

- after the implementation of the mongodb error middleware the order of the middlewares has been changed which caused our middleware to try to map unhandled errors to a httpError
- the mongodb error middleware broke when a mongodb error with an unknown code was thrown. I redirected those errors to be unhandled for now
- unhandled errors needed the statusCode of 500 assigned to them to be considered errors on the client side (how did I never notice this)

By investigating this issue I discovered a new one: https://github.com/ospin-web-dev/hambda/issues/299


lib/middleware/mapHandledErrorToProtocol.js
```diff
@@ -8,7 +8,7 @@ const httpHandler = (handler, next) => {
   }
   handler.response.body = message
   handler.response.statusCode = statusCode
-  return next()
+  return next(handler.error)
```
vdiezel: still don't now where the difference is but did it for consistency

lib/middleware/mapHandledErrorToProtocol.js
```diff
@@ -19,12 +19,15 @@ const defaultHandler = (handler, logger, next) => {
...
-      return defaultHandler(handler, logger, next)
+  if (handler.error.handled) {
+    switch (protocol) {
+      case 'http':
+        return httpHandler(handler, next)
+      default:
+        return defaultHandler(handler, logger, next)
+    }
   }
+  return next(handler.error)
```
vdiezel: unhandled error are now called last (the order used to be reverse), so I need this check here now

lib/middleware/mongoErrorToHandledError.js
```diff
@@ -12,7 +12,7 @@ module.exports = () => ({
   onError: (handler, next) => {
     const { error } = handler
 
-    if (isMongoError(error)) {
+    if (isMongoError(error) && (error.code in MONGO_ERROR_CODE_TO_HANDLED_ERROR)) {
```
vdiezel: considers only those errors handled which are in the map `MONGO_ERROR_CODE_HANDLED_ERROR`

lib/middleware/stringifyResponseBody.js
```diff
@@ -26,8 +26,9 @@ const errorMiddleware = (handler, next) => {
     })
   } else {
     handler.response.body = JSON.stringify({
-      message: handler.error.message,
+      message: handler.error.message || handler.error.errmsg,
```
vdiezel: error for unhandled mongo errors is on `handler.error.errmsg`

lib/middleware/stringifyResponseBody.js
```diff
@@ -26,8 +26,9 @@ const errorMiddleware = (handler, next) => {
...
     })
+    handler.response.statusCode = 500
```
vdiezel: needed this so unhandled errors are considered errors on the frontend. For some reason it's not automatically set (and it probably never was)

lib/middleware/stringifyResponseBody.js
vdiezel: Question: Should we provide any information for unhandled errors at all or better just return a generic message to the client? The error itself is still logged in CloudWatch, so we don't necessarily need this information on the clinet side, but it's probably more comfortable.

lib/middleware/mapHandledErrorToProtocol.js
DanielSeehausen: I interpreted it that calling next without passing the `handler.error` no longer bubbles up the error, and that passing it does!

lib/middleware/mongoErrorToHandledError.js
DanielSeehausen: good catch - this was an oversight on our part when implementing

lib/middleware/stringifyResponseBody.js
DanielSeehausen: I believe some people see this as a potential security flaw. 
I don't see an issue for us though. 
If we need to address it later maybe we could go the route of "on dev, return error message, otherwise don't"

lib/middleware/stringifyResponseBody.js
DanielSeehausen: I have a quick observation re: this middleware.

Now that things are becoming increasingly complex with our middleware, we should try to keep them functionally independent (and ignorant) of one another as much as possible. When conor and I were working in the middleware recently we (or at least I) found the coupling to be pretty difficult to follow and debug.

Should stringify response body have an onError action at all? It seems pretty pure as a middleware: if the response has a body, stringify it if it can be stringified.

Is the following assumption correct?: when this (or any) middleware is hit _after_ a handled error is handled in a middleware, it is no longer an error, and it shouldn't be hitting its onError in the case of a handled error.

That seems right and the original intent, but looking at this middleware's implementation it doesn't look like that is what is happening.

lib/middleware/mapHandledErrorToProtocol.js
DanielSeehausen: I left a comment in the middlewares mentioning this after conor and I worked on it. Was very strange and I couldn't identify why!

lib/middleware/stringifyResponseBody.js
vdiezel: splitting up by environment is a good idea!

lib/middleware/stringifyResponseBody.js
vdiezel: that is definitely a valid point. I just can't remeber if the error handler in the stringify body middleware was done because of a rush, lack of understanding or because I had too...
I create a new ticket for this one.

vdiezel: https://github.com/ospin-web-dev/hambda/issues/301

lib/middleware/mapHandledErrorToProtocol.js
vdiezel: I tried that and it doesn't seem to be the case. It seems more like it checks if there is an error present on the handler, and if so, call the onError middleware. At least behaviour never seemed to be altered when I switch between `next()` and `next(handler.error)`

ConorTighe1995: It looks kinda like ExpressJs in the sense if you want the callback to proceed you just call next() but passing a parameter will throw whats passed eg. next('throw this'). I've only seen this with ExpressJs but it must be a callback pattern for JS libs.

PR #303, opened 2020-03-19
fires the post confirmation hook only on signup, not on password reset

### LINKED ISSUE #302: Reset password error: mapErrorCodeToHTTPResponse
Screenshot from 2020-03-11 10-58-25

When attempting to reset my password I encountered this error once I received the validation code and then tried to login using my new password. Once I navigated back to the normal login screen and tried to login it worked fine.

vdiezel:
closes #302 

the issue was that the post confirmation trigger that is fired when the user sign ups for our application and calls our `user/create` endpoint is also fired when a user reset his/her password. There is no way to seperate them on AWS side. But you can tell the events apart using the triggerSource key in the lambda function. Thats the correct event:
Bildschirmfoto von 2020-03-19 19-51-06

The bad guy

Bildschirmfoto von 2020-03-19 20-10-33

And this is the result: 
Bildschirmfoto von 2020-03-19 19-59-29

already deployed to dev and prod.


services/user/src/user/create.js
```diff
@@ -7,6 +7,11 @@ const HTTPDefaultMiddleware = require('middleware/HTTPDefaultMiddleware')
 const User = require('./user')
 
 async function create(event, context) {
+
+  if (event.triggerSource !== 'PostConfirmation_ConfirmSignUp') {
+    context.succeed(event)
+  }
```
vdiezel: if the event is not the one we are looking for, return.

services/user/serverless.yml
```diff
@@ -63,8 +63,8 @@ resources:
           # EmailConfiguration:
           #   ReplyToEmailAddress: String
           #   SourceArn: String
-        EmailVerificationSubject: '(OSPIN) Your verification code'
-        EmailVerificationMessage: 'Welcome to OSPIN, Your verification code is {####}.'
+        EmailVerificationSubject: 'OSPIN Your verification code'
+        EmailVerificationMessage: 'The verification code for your request is {####}.'
```
O-Susanoo: You have gone a long way from Welcome to OSPIN to The verification code is , we can clearly see that ospin app made more mature :)

DanielSeehausen: lol

services/user/src/user/create.js
DanielSeehausen: I see `context.succeed`, but no return
is it needed or does `.succeed` take care of it?

image

services/user/src/user/create.js
```diff
@@ -7,6 +7,11 @@ const HTTPDefaultMiddleware = require('middleware/HTTPDefaultMiddleware')
 const User = require('./user')
 
 async function create(event, context) {
+
+  if (event.triggerSource !== 'PostConfirmation_ConfirmSignUp') {
+    context.succeed(event)
```
ConorTighe1995: Is `context` related to React context? aka useContext?

services/user/serverless.yml
ConorTighe1995: Do you guys support or will support localization? Or at least German? curious as we have so many German speakers that could translate the app.

services/user/src/user/create.js
vdiezel: actually, success is never called. 

Because the function is invoked by Cognito, its API expects a special kind of response which is `context.success` with the complete event as the argument.

By now `context.success` is actually deprecated as I discovered just now, and you can use the `callback` function: 
https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-pre-sign-up.html

I will fix this!

services/user/serverless.yml
vdiezel: thats not on our agenda currently - research is very international. Quite a few who bought our products in germany didn't even speak german.

services/user/src/user/create.js
vdiezel: no, its actually the now deprecated API to tell lambda when you want to return your result. I'll try to update this, using the `callback` API (see Daniel post further up)

services/user/serverless.yml
vdiezel: The sad thing about this email here is that again its the same one being send on post sign up and on request reset password events. If you want to have different ones, you have to create your own lambda functions to do that.

vdiezel:
Ok I tried to implement the `callback` function instead of the `context.succeed` and had a deja vu, going thorugh the same problems. When you use `callback`, you can't make the handler async, when you can make the handler async you have to make all our usual call with callbacks. I don't want to solve this now, so I leave it for now as it is.

PR #308, opened 2020-03-21
301 clean up stringify response body middleware

### LINKED ISSUE #301: clean up stringifyBody middleware
currenrtly, the stringifyBody middleware does more than it claims to, including different behavior when errors happen. This behaviour should be moved to the responsible error handlers if possible or abstracted into a new middleware if needed. See comments here: 

https://github.com/ospin-web-dev/hambda/pull/300

vdiezel:
closes #301 
 
Changed the execution flow of middleware a bit: 
I removed the middleware `logUnhandledErrors` completely and moved the logic for this to the protocol specific handlers to `mapErrorToProtocol` which I renamed to `handleErrorByProtocol` because it takes care of handled and unhandled errors now. This gives us now the freedom to care of unhandled errors in a protocol specific way if we ever need to. 

I couldn't get rid of the `onError` middleware in the `stringifyResponseBody` because you actually can't control the flow of middy using `next()` (of which we suspected it would call the next `before` or `after` middleware) or `next(handler.error)` (of which we thought it calls the next `onError` middleware). The bahaviour of middy is this: once it enter `onError` it proceeds always calling the next `onError` function, no matter what you pass into `next, as they mention in their docs: 

https://www.npmjs.com/package/middy#writing-a-middleware

> When a middleware handles the error and creates a response, the execution is still propagated to all the other error middlewares and they have a chance to update or replace the response as needed. At the end of the error middlewares sequence, the response is returned to the user.

So I just passed the same handler into `onError` as in `after` which solves the issue and makes `stringifyResponseBody` be single responsibility again.


DanielSeehausen:
> closes #301
> 
> Changed the execution flow of middleware a bit:
> I removed the middleware `logUnhandledErrors` completely and moved the logic for this to the protocol specific handlers to `mapErrorToProtocol` which I renamed to `handleErrorByProtocol` because it takes care of handled and unhandled errors now. This gives us now the freedom to care of unhandled errors in a protocol specific way if we ever need to.
> 
> I couldn't get rid of the `onError` middleware in the `stringifyResponseBody` because you actually can't control the flow of middy using `next()` (of which we suspected it would call the next `before` or `after` middleware) or `next(handler.error)` (of which we thought it calls the next `onError` middleware). The bahaviour of middy is this: once it enter `onError` it proceeds always calling the next `onError` function, no matter what you pass into `next, as they mention in their docs:
> 
> https://www.npmjs.com/package/middy#writing-a-middleware
> 
> > When a middleware handles the error and creates a response, the execution is still propagated to all the other error middlewares and they have a chance to update or replace the response as needed. At the end of the error middlewares sequence, the response is returned to the user.
> 
> So I just passed the same handler into `onError` as in `after` which solves the issue and makes `stringifyResponseBody` be single responsibility again.

very awesome. I certainly appreciate you taking the time to follow up to questions re: the middleware. This makes 100% sense

services/process/test/trigger/dataPointsToDB.test.js
```diff
@@ -60,9 +60,9 @@ describe('dataPointsToDb trigger', () => {
       500: { times: [], values: [] },
     }
 
-    savedDataPoints.forEach(({ times, values, sid }) => {
-      expect(times).toEqual(expectedStreamData[sid].times)
-      expect(values).toEqual(expectedStreamData[sid].values)
+    savedDataPoints.forEach(({ times, values, streamId }) => {
+      expect(times).toEqual(expectedStreamData[streamId].times)
+      expect(values).toEqual(expectedStreamData[streamId].values)
     })
```
DanielSeehausen: unsure if you meant for this to be in this PR or not, as it wasn't mentioned in the PR comment

No complaints if you are just sneaking it in - just wanted to point it out if _it_ is the one who snuck by _you_ as a stray commit

vdiezel: no, it was intended to be here, it the second time changes which breaking tests were merged into master unseen and I still don't know how that happened.

PR #309, opened 2020-03-21
restricts the topics a user can subscribe to

### LINKED ISSUE #34: as a user I can only subscribe and connect to IoT for topics that I have access to
Even though a user can receive messages only from devices the user has access to, the user still can subscribe and connect to topics of other devices (even though he will never be able to actually receive anything - weird, I know). 

PS: I couldn't  achieve preventing unauthorized subscriptions without breaking the allowed subscriptions - Viktor

vdiezel:
closes #34 

this is a minor security risks, because user so far were able to subscribe to any topic they want but couldn't actually receive messages on that topic:

https://docs.aws.amazon.com/iot/latest/developerguide/iot-policy-actions.html

Main usage is to be able to prevent sending messages on the server side without cancelling the subscription of the client. 

updated all existing policies.


vdiezel:
It took a while until the policies propagated and then it stopped working - getting back to work...

PR #311, opened 2020-03-22
adds endTime key on data chunks

### LINKED ISSUE #310: save the end time on the data chunks to allow to query data by it.
Currently, only the startTime of a data chunk is stored on it, but we also need the endTime to be able to query data properly.

vdiezel:
closes #310 

adds the endTime key to data chunks.


services/cortex/models/Evacuator.js
```diff
@@ -15,8 +15,7 @@ class Evacuator {
       processId,
       streamId,
       startTime: streamData.x[0],
-      sid: streamId,
-      t: streamData.x[0],
```
vdiezel: this keys were deprecated in the last migration

DanielSeehausen:
used the interface to rebase so your tests up to date. Was clean.

Issue #200, opened 2020-01-21
change owners group to admins


DanielSeehausen:
and double check front end is ready for the change
might need ticket for reactor as well

PR #315, opened 2020-03-23
fix test reference

DanielSeehausen:
missed a reference!
related to #200


DanielSeehausen:
> Guess I should have `Ag`ed too before approving the original PR. I imagen you left this one in on purpose to test my accuracy and I failed.

i wish I were this cunning
but if it helps drive you viktor, this can be our truth

PR #317, opened 2020-03-24
Add configurator service

### LINKED ISSUE #316: Add configurator service in hamda for sending emails

O-Susanoo:
This pr add the ability to send email that contain image that user create while using the configurator
closes #316


lib/models/Configurator/Configurator.js
```diff
@@ -0,0 +1,45 @@
...
+
+  static buildEmail(userName, deviceName) {
+    return `
+    <!DOCTYPE html>
+    <html>
+      <head></head>
+      <body><h1>Hi ${userName} here is your device ${deviceName}</h1></body>
+    </html>
+    `
+  }
```
O-Susanoo: this stuff will be changed in future once marketing provide me with email they want to send , because nothing is being stored into the database so it must be manually inserted at least for now

services/configurator/src/configurator/create.js
```diff
@@ -0,0 +1,21 @@
...
+
+async function create(event) {
+  const { body: requestData } = event
+
+  validateCreateConfigurator(requestData)
+  const data = {
+    userName: requestData.userName,
+    deviceName: requestData.deviceName,
+    userEmail: requestData.userEmail,
+    configurator: requestData.configurator,
```
O-Susanoo: this one is never being used but it kept for future in case we want to store users the configurator they create it will mainly be a description of the module of the device

services/configurator/src/configurator/create.js
```diff
@@ -0,0 +1,21 @@
...
+const HTTPDefaultMiddleware = require('middleware/HTTPDefaultMiddleware')
+const Configurator = require('models/Configurator/Configurator')
+const { validateCreateConfigurator } = require('../joi/configurator')
+
+async function create(event) {
+  const { body: requestData } = event
+
+  validateCreateConfigurator(requestData)
+  const data = {
+    userName: requestData.userName,
```
vdiezel: I'm wondering if the user name is really necessary - I would wonder as somebody who uses the configurator what its being used for. 
I see its being used in the email body, but you could just ommit it there too.

Same applies for the device name - do we really need it?

O-Susanoo: well this stuff was requested by them i don't know how they are going to use it , but if they are preset they should be inserted in the email if no maybe another template ? again i will need to check with Felix to see what's his input on this one

services/configurator/src/configurator/create.js
DanielSeehausen: I'd recommend destroying it until it is used! potential YAGNI

services/configurator/src/joi/configurator.js
```diff
@@ -0,0 +1,14 @@
+const Joi = require('@hapi/joi')
+const { createSchemaAssertion } = require('validation-lib')
+
+const validateCreateConfigurator = createSchemaAssertion(Joi.object({
+  userName: Joi.string().max(80).optional(),
+  deviceName: Joi.string().max(80).optional(),
+  userEmail: Joi.string().email().required(),
+  configurator: Joi.object().optional(),
```
ConorTighe1995, line 8: whats configurator here?

services/configurator/src/configurator/create.js
ConorTighe1995: I made a comment but resolved it, this basically answers what I asked.

PR #333, opened 2020-03-26
318 migrate process data to track pinned

### LINKED ISSUE #318: PIN 1: Migrate process data to track `pinned`
#### Deliverables
- [x] update process create to add pinned false and use a JOI schema for creation instead of written in line
- [x] double check process clone
- [x] add `pinned: false` to all processes (dev + prod) via a script
- [x] update tests to pass in `pinned` booleans and more
- [x] joi schema on update process needs to accept the new pinned key and restrict to boolean

As written above, this might have the opportunity to break - consider a way to ensure it can't break in-between deploying this and migrating it, or the other way around. Reactor object mapper will also probably discard `pin`...unsure if that will make a difference on edit/create calls.

DanielSeehausen:
for the new Logger, I almost didn't write any tests...then found so many bugs while doing so:

image

definitely recommend going through commit wise! 

closes #318


DanielSeehausen:
I have no idea how the `scripts/EditResourceAccess.js` coverage went down...I struggle reasing codecovs app sometimes

DanielSeehausen:
image

image

scripts/AddField.js
```diff
@@ -0,0 +1,21 @@
...
+
+class AddField {
+
+  static toCollection(collection, fieldName, fieldValue) {
+    const setQuery = {}
+    setQuery[fieldName] = fieldValue
+
+    return collection.updateMany(
+      {},
+      { $set: setQuery },
```
vdiezel: you could do this: `const setQuery = { [fieldName]: fieldvalue }`
and because its being used only once, you could actually skip the whole initialization and add it to immediately in line 13

test/scripts/Logger.test.js
```diff
@@ -0,0 +1,72 @@
...
+  describe('watchConsole', () => {
+    it('logs everything passed to console.log and console.warn', async () => {
+      const logSpy = jest.spyOn(console, 'log').mockImplementation()
+      const warnSpy = jest.spyOn(console, 'warn').mockImplementation()
+
+      Logger.logTo('watchConsoleLogTest')
+      Logger.watchConsole()
+
+      const logLine = 'Hell, I can get you a toe by 3 o\'clock this afternoon... with nail polish'
+      const warnLine = 'Look, Larry. Have you ever heard of Vietnam?'
```
vdiezel: you are entering a world of pain, son

services/process/src/process/create.js
```diff
@@ -9,8 +9,19 @@ const Process = require('./process')
 async function create(event) {
   const requestData = {
     requestingUserId: event.requestingUserId,
-    deviceId: event.queryStringParameters.deviceId, // TODO: should this be a path parameter?
-    data: event.body
+    deviceId: event.queryStringParameters.deviceId,
+    data: {
+      name: event.body.name,
+      deviceId: event.queryStringParameters.deviceId, // TODO: should this be a path parameter, understanding device is required (i.e. is this really create device process)?
```
vdiezel: As we dream of processes and devices being independent resources, this should really be in the event body I guess.

services/process/src/process/create.js
```diff
@@ -9,8 +9,19 @@ const Process = require('./process')
 async function create(event) {
   const requestData = {
     requestingUserId: event.requestingUserId,
-    deviceId: event.queryStringParameters.deviceId, // TODO: should this be a path parameter?
-    data: event.body
+    deviceId: event.queryStringParameters.deviceId,
```
vdiezel: deviceId is now present on `requestData` object and withtin `requestData.data`. I would get rid of one, probably the top level one.

services/process/src/joi/process.js
```diff
@@ -15,15 +16,7 @@ const getPublic = createSchemaAssertion(Joi.object({
...
-  data: Joi.object({
-    name: Joi.string().min(1).max(45).required(),
-    entryPhaseId: Joi.number().strict().required(),
-    description: Joi.object().required(),
-    deviceDescription: Joi.array().required(),
-    deviceName: Joi.string().strict(),
-    cloneHistory: Joi.array(),
-    comment: Joi.string().allow('').max(250).strict(),
-  }).required(),
+  data: Process.SCHEMA.required(),
```
vdiezel: the `name` key seems to be missiing on your schema.

services/process/src/process/process.js
```diff
@@ -32,6 +32,7 @@ class Process {
       createdAt: new Date().getTime(),
       createdBy: requestingUserId,
```
vdiezel: when you set the `createdBy` property now on the data object, you should also using it from there, instead of assigning the `requestingUserId` here again.

services/process/src/process/process.js
```diff
@@ -236,6 +237,7 @@ class Process {
     process.access = createDefaultGroups('process', requestingUserId);
     process.createdAt = new Date().getTime();
     process.id = uuidv4()
+    process.pinned = false
     process.startedBy = null
```
vdiezel: do we ever set the key `startedBy` anywhere? If not, could you make a ticket to set it?

scripts/AddField.js
```diff
@@ -0,0 +1,21 @@
...
+
+
+class AddField {
+
+  static toCollection(collection, fieldName, fieldValue) {
+    const setQuery = {}
+    setQuery[fieldName] = fieldValue
+
+    return collection.updateMany(
+      {},
```
ConorTighe1995: Why is an empty object being passed to the update function?

test/scripts/Logger.test.js
```diff
@@ -0,0 +1,72 @@
...
+
+  describe('watchConsole', () => {
+    it('logs everything passed to console.log and console.warn', async () => {
+      const logSpy = jest.spyOn(console, 'log').mockImplementation()
+      const warnSpy = jest.spyOn(console, 'warn').mockImplementation()
+
+      Logger.logTo('watchConsoleLogTest')
+      Logger.watchConsole()
+
+      const logLine = 'Hell, I can get you a toe by 3 o\'clock this afternoon... with nail polish'
```
ConorTighe1995: liking the toe story telling, looking forward to your next to PR to see where this narrative takes us

services/process/src/process/create.js
DanielSeehausen: 👍

scripts/AddField.js
DanielSeehausen: this is a good tip - I will update but not in this pr to avoid re-review

scripts/AddField.js
DanielSeehausen: good 👁 @ConorTighe1995 

the first param in mongo calls like this is the `filter` param
in non-single resource calls (i.e. update**Many**), passing an empty object is like saying `filter: all`

Here is a snapshot from mongo docs

image

DanielSeehausen:
> Looks good man, left a question but thats it.

Y

DanielSeehausen:
oops closed and reopened

test/scripts/Logger.test.js
DanielSeehausen: `test/lib/models/EventRevoker.test.js` has the prequel to this same story line :))

scripts/AddField.js
DanielSeehausen: scratch that ill update now

services/process/src/process/process.js
```diff
@@ -32,6 +32,7 @@ class Process {
       createdAt: new Date().getTime(),
       createdBy: requestingUserId,
       name: data.name,
+      pinned: data.pinned,
```
O-Susanoo: it seem to me that  pinned is optional at least from the joi schema is that normal or should we add this 
``` 
pinned : data.pinned || false 
```

test/scripts/Logger.test.js
O-Susanoo: Hahaha :)

services/process/src/process/create.js
DanielSeehausen: made a ticket

services/process/src/process/process.js
DanielSeehausen: good 👁

services/process/src/process/process.js
DanielSeehausen: in device `startProcess` we set it - 
for example, this is how we resolve the info on a running device re: who started it

vdiezel: :+1:

services/process/src/process/process.js
DanielSeehausen: good 👁 @O-Susanoo 
pinned is optional in the update joi

this is the create method, and the JOI create assertion (for the lambda function) extends Process.SCHEMA, which defaults to `false` here. Ideally, we should never get to this point without having a value for pinned!

scripts/AddField.js
ConorTighe1995: Ok cool thanks

PR #354, opened 2020-03-30
318 cleanup

### LINKED ISSUE #355: process.getWithDevice bug
`getWithDevice` query for processes is attempting to delete a key that does not exist.
`delete ... .device` should have been `devices`.

That bug revealed another issue, in that cloning processes was unintentionally persisting objects with a `devices` key on them (because the clone function made use of this query).

DanielSeehausen:
i was abusing `static`

this cleans it up to make logger an instance and less user error prone
and hopefully improves self documentation on the usage

updates tests as well

just tested this all with the pinned stuff and is working as intended:

image

Minor note re: 318 - the npm commands to run scripts in certain environments won't work until our env files are updated to not be yml files but instead be like so (which is coming in PRs for hambda CD so no worries):

`STAGE='prod'`

closes #355


services/process/src/process/process.js
```diff
@@ -105,7 +105,7 @@ class Process {
 
     const processWithDevice = processWithDeviceArray[0]
     const device = processWithDevice.devices[0]
-    delete processWithDevice.device
+    delete processWithDevice.devices
```
DanielSeehausen: TDD will prevent bugs like this!
How this bug was manifesting itself certainly spent more time to find than writing test coverage :(

.gitignore
```diff
@@ -34,9 +34,9 @@ tmp
 .lock-wscript
 .env
 env.yml
-env.dev.yml
-env.prod.yml
-env.staging.yml
+env.dev.*
+env.prod.*
+env.staging.*
```
vdiezel: ill make those probably dot-files

PR #360, opened 2020-03-31
New deployment strategy for hambda

### LINKED ISSUE #328: Hambda CD 2: CI/CD for staging branch/env
## Deliverables
- [ ] add add ci/cd for staging branch through CircleCI for _lambda services only_ (excludes cortex)
  - merging to staging auto deploys all services to staging env on tests green
  - the pipeline alerts team in circleCI channel (if its easy, add names/github handles or whatever is passed to circle ci to alerts)

vdiezel:
closes #328 

Modifies circleCI config to CD our code using the serverless framework and changes the wa local development happens

### Why isn't the deployment of the seperate services parallel?

The only way I managed to make this steps parallel was putting the deployment of every service into a seperate job. But this way I needed to attach the workspace for every service (and also install serverless package globally, which maybe COULD be cached but the resources on the internet are all confusing). This took 1-1.5 minutes per service which eats up the advantage of parallelizing them (it would save 1-2 minutes in total, but eats our credits much quicker). Also because we use a free plan, parallel jobs are often queed and do not execute in parallel. 

Other approaches I tried: 

#### Pushing jobs into the background
https://discuss.circleci.com/t/parallel-commands/13468/10

##### using `background: true`
This does not work, as the execution flow does not wait for background job to finish. 

##### using `run: | `
This does not work, as this does not spawn a new shell that we need for every deployment

#### using the predefined  `deploy` step
https://circleci.com/docs/2.0/configuration-reference/#deploy

This does not do what we need. `deploy` actually waits for all parallel jobs on all nodes (in case of parallelism) to finish, before starting to deploy. What we need is parallel deployment itself.

### What do I need to know for local development

First of all the yml-env-files are now replaced with `.env.{stage}` files and contain the env vars in a slightly different format. I will upload them soon. 

Second, because the serverless files now expect environmental variables defined in their execution context, I had to wrap `sls deploy` into an `npm run` command that injects the environmental variables. I couldn't solve this using the `serverless-scriptable-plugin` because  there is no life cycle event that happens early enough to inject them before sls tries to compile the `serverless.yml`.
__Which means now you have to deploy using `npm run deploy`!__ You still can pass any parameters to it, using this syntax `npm run deploy -- --function myFunctionName`
This always deploys to dev. If you want to deploy somewhere else, you can manually export the desired env file and then run any sls command. I didn't write scripts in the package.json because we don't want to encourage this. But could be added very easily...

## How does a user from production get access to staging API gateways? 

There were to ways to solve this: 
The insecure, automated way: hard code the API gateways for staging in the user services `serverless.yml`. But this way also users from dev would get access to `staging` which I didn't like. 

The secure, manual way: I created a new policy on AWS called `prod_API_access_to_staging` and attached it to the production user. New deploys should not overwrite this (tested it and it seems to work). The disadvantage is that whenever you want to add a new service, you need to remember to modify this policy.


services/event/serverless.yml
```diff
@@ -7,8 +7,8 @@ plugins:
 
 provider:
   name: aws
+  stage: ${{env:STAGE}}
```
vdiezel: stage is not passed in as a parameter to the deployment command but derived from the set environemtal vars.

services/event/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "scripts": {
     "seed-lib-files": "cp -rf ../../lib/* ./node_modules/",
+    "deploy": "export $(cat ../../.env.dev | sed 's/#.*//g' | xargs) && sls deploy",
```
ConorTighe1995, line 4: noiceeee

services/user/serverless.yml
```diff
@@ -205,7 +205,7 @@ resources:
       Type: AWS::IAM::Role
       Properties:
         Path: /
-        RoleName: ${{self:service}}-${{self:provider.stage}}-lambda-cloudWatch
+        RoleName: ${{self:service}}-${{env:STAGE}}-lambda-cloudWatch
```
ConorTighe1995, line 208: this for logs?

vdiezel: the role - yes.For every lambda function you can defined with which role it should be executed . This way you have maximum control over permissions by giving access only to those services the lambda function needs. The default role is the role mentioned here that allows creating logs for CloudWatch

vdiezel:
> I've approved it but left some comments, in the future we should document a lot of this in confluence. I have no exposure to Circle CI and seems to be a lot in it. Good stuff though!

CircleCI has a really good documentation luckily - the real dread is AWS

ConorTighe1995:
> > I've approved it but left some comments, in the future we should document a lot of this in confluence. I have no exposure to Circle CI and seems to be a lot in it. Good stuff though!
> 
> CircleCI has a really good documentation luckily - the real dread is AWS

Nice ill try take some time to run over it

PR #370, opened 2020-04-01
Fix Circle Ci Failing Tests

### LINKED ISSUE #364: Hambda test suite passing in circleCI even when tests are failing
locate and fix issue!

here as an example of when tests fail, yet CircleCI reports green :(

Believe I may have introduced this when running the tests in parallel?

O-Susanoo:
closes #364


O-Susanoo:
I used this command multiple times on this branch
```
time npm run testAllWithCoverage
```
so that i can mimic the same thing on circle ci
real	0m59.620s
real	0m47.704s
real	0m52.086s

so i guess we can assume a 53 seconds average 

on circle ci this take 6:30 , 6:08 , 6:56, so an average of 6:30

on the other hand when it's async
it's 
real	0m16.250s
real	0m13.660s
real	0m15.645s

an average of 14s so it's almost 4 times faster locally

on circle ci it's 2:06 , 1:20 , 1:20 so an average of 1,5 minutes
but that without cortex because cortex is not included now on master (bug) but i think it's also a factor of 4 which is not negligable

PR #372, opened 2020-04-02
371 view running process

### LINKED ISSUE #371: Check if user has access to view a process on device and return results to REACTor
In order for REACTor to know if it should allow a user to view a particular process running on a device we need to determine if they have access to view the processes and if they are an admin. The backend should return a hasAccess value when loading the devices page to allow the front end to handle what processes they can view from the devices displayed. The backend should not expose the roles assigned to the users.

ConorTighe1995:
The point of theses changes was to allow REACTor to distinguish between who has access to virw a process running on a device. We provide this by checking if the user has access to the resource and is an admin, then we use these checks to validate that the user has access to process. We also did some refactoring in the validation lib using the introduced access right functions. Now when the get public function is called we can expose a who has access to view without exposing the roles in the backend.

closes #371


DanielSeehausen:
is the first commit (new deployment strategy for hambda) supposed to be in this PR?

lib/access-rights-lib.js
```diff
@@ -231,3 +231,15 @@ exports.getAllUserIdsFromResource = resource => (
...
+  resource.access.some(resourceAccessGroup => resourceAccessGroup.users.includes(userId))
+)
+
+const userIsAdminOnResource = (resource, userId) => (
+  resource.access.some(resourceAccessGroup => resourceAccessGroup.users.includes(userId) && resourceAccessGroup.name === 'admins')
+)
+
+exports.userHasProcessAccess = (process, device, userId) => (
+  (isUserOnResource(process, userId) || userIsAdminOnResource(device, userId))
+)
```
DanielSeehausen: this access right lib is going to keep moving in the direction of something that can be shared between any resource with the `access` interface (I suppose you'd call any object with the `access` a duck type in regards to access operations)

Understanding that, we might want to keep the naming generic, as the functions that compose this function are already generic themselves!

ConorTighe1995: so `hasProcessAccess` would be more suitable?

lib/access-rights-lib.js
```diff
@@ -231,3 +231,15 @@ exports.getAllUserIdsFromResource = resource => (
...
   ]), [])
 )
+
+const isUserOnResource = (resource, userId) => (
+  resource.access.some(resourceAccessGroup => resourceAccessGroup.users.includes(userId))
+)
+
+const userIsAdminOnResource = (resource, userId) => (
+  resource.access.some(resourceAccessGroup => resourceAccessGroup.users.includes(userId) && resourceAccessGroup.name === 'admins')
+)
```
DanielSeehausen, line 242: the method below this is composed of two distinct parts, and this looks like a good candidate as well:

```js
  return getResourceAdmins(resource) // this method will 100% be used elsewhere, and there should only be interface for getting admins of a resource. This type of access code is repeated a lot, and target for a refactor at some point
                .includes(userId)
```

services/process/src/process/process.js
```diff
@@ -124,16 +138,22 @@ class Process {
 
   async getPublic({ requestedProcessId, requestingUserId }) {
     const requestedProcess = await this.processesDB.findOne({ id: requestedProcessId })
-
     assert(() => canFind(requestedProcess, `Process ${requestedProcessId}`))
 
-    const serializedProcess = serializePublicProcess(requestedProcess, requestingUserId)
+    const device = await this.devicesDB.findOne({ id: requestedProcess.deviceId })
+    assert(() => canFind(device, `Device ${requestedProcess.deviceId}`))
```
DanielSeehausen, line 145: I'm unsure the public fetch of a process should fail if it can't find an associated device - we might be taking on tech debt doing so. It appears we are checking if the device exists because we want to not continue the function because device is used within it. My concern is, that fail fast is being used for an execution convenience, and not because it serves the `getPublic` method's intended result, even though it has the power to 400 the call. 

These changes seem to turn this endpoint into something closer to a:
`GET device/:id/process/:id/public`
because it becomes reliant on the device (it will blow up if device is not found)
or perhaps even more honest is to include user, which perhaps should have been included the first time seeing as `requestingUserId` is passed in:
`GET user/:id/device/:id/process/:id/public`
I believe that brings us closer to the correct route, as we are saying:
Get the public data available, for a process that is tied to a specific device, which a specific user is authorized to receive. This route rambling is just something to noodle on - not sure any change should happen here before we make these proper Models, which will largely take care of this public/private concept by ensuring the models don't own authorization (then different routes can be used, instead of methods)

Instead of going through all that headache above, an alternative way out would be to: 
ensure `device` is an optional parameter on `userHasProcessAccess`. If not passed, the device authorization fallback function within `userHasProcessAccess` can't be true
or:
separate the process and the device access calls and use them both here in this method:
```js
const userIsDeviceAdmin = device && userIsAdminOnResource(device, userid)
const userHasAccess = isUserOnResource(process, userId) || userIsDeviceAdmin
```

This would make it easier to not need to fail on not finding the device.

services/process/src/process/process.js
```diff
@@ -124,16 +138,22 @@ class Process {
...
+
+    const userHasAccess = userHasProcessAccess(requestedProcess, device, requestingUserId)
+
+    
+
+    const serializedProcess = serializePublicProcess({ ...requestedProcess, hasAccess: userHasAccess })
```
DanielSeehausen, line 149: since this is no longer just about a process, and tied to a user, i'd like to request it becomes a `serializePublicUserProcess` or something similar for the time being.

DanielSeehausen: (this would also require nexus/reactor change)

lib/validation-lib.js
```diff
@@ -209,7 +209,7 @@ const isResourceAdmin = (resource, userId) => {
 
 exports.isResourceAdmin = isResourceAdmin
 
-exports.userHasProcessAccess = (device, process, userId, action) => (
+exports.userHasProcessActionAccess = (device, process, userId, action) => (
```
DanielSeehausen, line 213: 👍

services/process/src/process/process.js
ConorTighe1995: Im just looking at refactoring now, do you want me to replace the asserts here and add a new function `isUserOnResource?` to access -rights-libs?

ConorTighe1995: So something like this here?

```
async getPublic({ requestedProcessId, requestingUserId }) {
    const requestedProcess = await this.processesDB.findOne({ id: requestedProcessId })
    assert(() => canFind(requestedProcess, `Process ${requestedProcessId}`))

    const device = await this.devicesDB.findOne({ id: requestedProcess.deviceId })
    assert(() => canFind(device, `Device ${requestedProcess.deviceId}`))

    const userIsDeviceAdmin = device && userIsAdminOnResource(device, requestingUserId)
    const userHasAccess = userIsPresentOnResource(process, requestingUserId) || userIsDeviceAdmin

    const serializedProcess = serializePublicProcess({ ...requestedProcess, hasAccess: userHasAccess })

    return serializedProcess
  }
```

vdiezel: We went through the same thoughts when we paired on this. Because currently every process is bound to a device (and this will probably remain for quite a while) we didn't bother to much about where to put the check for access to the process. But your points are valid.

services/process/src/process/public/get.js
```diff
@@ -13,12 +25,25 @@ async function get(event) {
...
+  const { requestingUserId, requestedProcessId } = requestData
   const db = await mongo.getDB()
   const process = new Process(db)
+  const device = new Device(db)
+
+  const requestedProcess = await process.find(requestedProcessId)
+  assert(() => canFind(requestedProcess, `Process ${requestedProcessId}`))
+
+  const requestedDevice = await device.find(requestedProcess.deviceId) // find on device
+  assert(() => canFind(requestedDevice, `Device ${requestedProcess.deviceId}`))
```
DanielSeehausen, line 37: only one more minor change request
it might not be as clear as we could have made it from the naming, but `assert` will actually blow up in a specific way and then the error will go straight to middleware handling (the rest of this function after line 37 here wouldn't execute at all if that assert failed)

It might make more sense if we don't fail on the device not being found. The device being found is only consequential for the `userHasAccess` stuff, but if its not found, that `userHasAccess` stuff should ultimately go to false, and the requestor should still get this public process payload back because the finding of the process itself did not fail (on line 34).

as things are now, with this assertion in place, (and if it were to fail), we wouldn't need the `device &&` shortcircuit on line 39 because the assertion would have ended execution. Once we remove the assertion blow up, then that `device &&` will be doing something for us. 

Other than that, everything looks great. Let me know if you have any other questions or if I am not understanding this fully. I will re-approve once ready right away if you slack me!

DanielSeehausen:
looks great! I approved. Thank you for the patience with the PR reviews

PR #388, opened 2020-04-07
Implements a new service for event triggers

### LINKED ISSUE #382: abstract event triggers on cortex to a new service instead and connect the new service to redis
Because the redis instance we have runs remotely in the AWS cloud, the event triggers can be abstracted into a new service with lambda functions again. 

Further, the new service needs to be connected to the redis instance instead of utilizing the local cache (as it was doing on the cortex instance). This will require following steps: 

- create a new `eventTrigger` service and migrate the code there
- integrate the lambda functions and redis into the same VPC (mandatory)
- because using a vpc is mandatory here, the internet access of lambda will be restricted. In order to connect to mongo we need to integrate vpc access to Mongo Atlas, too
- implement a solution for process `update` hook to invalidate cached processes

vdiezel:
implements a new service named `eventTrigger`
closes #382 

The files `EventTriggerAPI` and the `publishTriggeredEvents.test.js` were actually already reviewed when I implemented cortex. So you can skip them if you want to - I did minor changes to them to address the implementation of redis. 

Rest is only helpers functions and configuration. 
env files don't need any adjustment.


lib/redisClient.js
```diff
@@ -0,0 +1,25 @@
...
+  redisClient.get = promisify(redisClient.get)
+  redisClient.keys = promisify(redisClient.keys)
+  redisClient.del = promisify(redisClient.del)
+
+  return redisClient
+}
+
+module.exports = global.process.env.NODE_ENV !== 'test'
+  ? createRedisClient()
+  : new CacheServiceSpoof()
```
ConorTighe1995, line 25: Is this meant to fill in for redis if the client fails to connect?

services/eventTrigger/src/eventTrigger/EventTriggerAPI.js
```diff
@@ -16,7 +16,10 @@ class EventTriggerAPI {
     return 1800000 // ms = 30 mins
   }
 
-  static async getProcess(processId) {
+  // to finalize the cache inmplementation we need to write
+  // a cache invalidation for processes first - for now we always load from the database
+
+  static async getProcessWithCacheCheck(processId) {
     const cachedProcessDataString = await this.redis.get(processId)
     let process = cachedProcessDataString
```
ConorTighe1995, line 24: could this be a const?

vdiezel:
going to create a new fresh one!

PR #398, opened 2020-04-09
V2 382 event trigger api

### LINKED ISSUE #382: abstract event triggers on cortex to a new service instead and connect the new service to redis

vdiezel:
closes #382 

adds a new service called `eventTriggers`
I had to add a bit of code that duplicates front end code - we will solve this later when we have some remote code delivery.

Please update your .env files with the `REDIS_PORT` and `REDIS_URL` env vars in hambda. Ive updated the pinned files in slack organ-farm. 

For review: 
The line count is not representative of the PR because most of it is code copied from cortex / duplicate of frontend code. Most of the files were reviewed already, I commented those. 
The meat of the PR is the `EventTriggerAPI.js` - please  review the final version here. Rest can be reviewed by commit.


services/eventTrigger/src/eventTrigger/EventTriggerAPI.js
```diff
@@ -0,0 +1,142 @@
+const Connectable = require('models/Connectable')
```
vdiezel, line 1: this file has been reviewed when I created cortex - it weill go thorugh some changes in the commits - best to reveiw the final version here.

services/eventTrigger/test/publishTriggeredEvents.test.js
```diff
@@ -0,0 +1,339 @@
+const Seeder = require('test-helpers/seeder/Seeder')
```
vdiezel, line 1: this file also has been reviewed already when creating cortex

services/cortex/app.js
```diff
@@ -4,9 +4,9 @@ const MQTTEventHandler = require('./models/MQTTEventHandler')
 const mongo = require('./lib/mongodb-lib')
 const CacheManager = require('./models/CacheManager')
 
-mongo.getDB().then(db => {
+const redis = require('./startup/redis')
```
vdiezel, line 7: this is not necessary here and will be removed.

services/eventTrigger/serverless.yml
```diff
@@ -16,9 +16,8 @@ provider:
   vpc:
     securityGroupIds:
       - sg-15129179
-
     subnetIds:
-      - subnet-1d131c76
+      - subnet-3287fd4f
```
vdiezel, line 20: this subnet has access to the AWS IoT broker via a NAT Gateway

lib/models/UnitConverter/UnitConverter.js
```diff
@@ -0,0 +1,154 @@
+/* Unit-Converter:
```
vdiezel, line 1: this code is a duplicate of our frontend unit conversion api

services/eventTrigger/src/eventTrigger/EventTriggerAPI.js
```diff
@@ -106,19 +103,21 @@ class EventTriggerAPI {
   }
 
   static mapTriggersToEvents(eventTriggers, process) {
+
+    const device = this.devicesDB.findOne({ id: process.deviceId })
```
vdiezel: I need the device to get the correct displayed sream name and customer adjusted stream units

services/eventTrigger/src/eventTrigger/models/OutOfBoundaryTrigger.js
```diff
@@ -8,9 +9,28 @@ class OutOfBoundaryTrigger extends EventTrigger {
...
+
+    const streamNameAlias = device.streamNameAliases.find(entry => entry.streamId === this.streamId)
+    const displayedStreamName = streamNameAlias ? streamNameAlias.alias : stream.name
+
+    // convert to unit
+
+    const streamConfig = device.config.streams.find(entry => entry.streamId === this.streamId)
+    const displayedUnit = streamConfig && streamConfig.defaultUnit
+      ? streamConfig.defaultUnit
+      : stream.unit
```
vdiezel: This code is also a duplicate of frontend code, will be abstarcted once we can share code between repos.

lib/test-helpers/CacheServiceSpoof.js
```diff
@@ -0,0 +1,30 @@
+class CacheServiceSpoof {
```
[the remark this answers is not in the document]
vdiezel: NO REVIEW REQUIRED: copy from cortex

services/eventTrigger/src/eventTrigger/models/DataUtils.js
```diff
@@ -0,0 +1,14 @@
+class DataUtils {
```
vdiezel, line 1: NO REVIEW REQUIED: copy from cortex

lib/models/Event/Event.js
```diff
@@ -38,6 +38,7 @@ class Event {
       deviceId: Joi.string().pattern(UUIDV4_REGEX).allow(null),
       processId: Joi.string().pattern(UUIDV4_REGEX).allow(null).default(null),
       createdAt: Joi.number().integer().min(0).required(), // Unix time
+      streamId: Joi.string().allow(null).default(null),
```
O-Susanoo, line 41: i'm pretty sure there is a good reason why a streamId can be null , but just mentioning it here in case it wasn't meant to be

lib/models/Process/Process.js
```diff
@@ -44,10 +45,12 @@ class Process {
     return downStreamIds.reduce((acc, streamId) => {
       const streamEventTriggers = phaseDownStreams[streamId].eventTriggers
 
+      const linkedUpstream = Streams.getTrackingUpstream(process.deviceDescription, streamId)
+
```
O-Susanoo, line 49: i think this one should go inside the if since it's only used there ?

services/eventTrigger/serverless.yml
O-Susanoo: well i guess you can add on your resume a new language , beside english and russian it's called serverless and probably 5 people speak it in the world

services/eventTrigger/src/eventTrigger/EventTriggerAPI.js
```diff
@@ -0,0 +1,147 @@
...
+    }
+  }
+
+  static async updateEventTriggerTimestamps(triggers, cachedTriggerData, processId) {
+    triggers.forEach(trigger => {
+      const { streamId, type } = trigger
+      this.updateEvent(cachedTriggerData, streamId, type)
+    })
+
+    await this.redis.set(processId + this.CACHE_SUFFIX, JSON.stringify(cachedTriggerData))
```
O-Susanoo, line 102: i'm not sure if what's i'm going to say is true or not but is would this code really work since forEach is async it will directly jump here thus not updating the events ?

services/eventTrigger/src/eventTrigger/models/OutOfBoundaryTrigger.js
```diff
@@ -0,0 +1,55 @@
...
+    super(unspecificData)
+    this.limit = limit
+  }
+
+  createEventMessage(pos, data, device) {
+    const stream = device.streams.find(deviceStream => deviceStream.id === this.streamId)
+
+    // find displayed stream name
+
+    const streamNameAlias = device.streamNameAliases.find(entry => entry.streamId === this.streamId)
```
O-Susanoo, line 17: is there a possibility that both of these can't be found

O-Susanoo:
Nice :+1:  jobe as always God grace on earth

lib/models/Process/Process.js
```diff
@@ -44,10 +45,12 @@ class Process {
     return downStreamIds.reduce((acc, streamId) => {
       const streamEventTriggers = phaseDownStreams[streamId].eventTriggers
```
ConorTighe1995, line 46: Could just deconstruct the event triggers

lib/models/Streams/Streams.js
```diff
@@ -0,0 +1,12 @@
+class Streams {
+  static getTrackingUpstream(streams, downstreamId) {
+    return streams
+      .filter(stream => stream.type === 'upstream')
+      .filter(stream => stream.relatedDownstreams && stream.relatedDownstreams.length)
+      .find(stream => stream.relatedDownstreams.some(relation => ((
+        relation.streamId === downstreamId || relation.streamID === downstreamId)
```
ConorTighe1995, line 7: There is a lot going on here, maybe this could be broken into functions to describe whats happening, one for filter and one for find and store the results in consts then return the tracked upstream/upstreams, also `relation.streamId === downstreamId || relation.streamID === downstreamId`??? why is there duplicated steam ids on this?

vdiezel:
> is there a possibility that both of these can't be found or even the first one because it's being actively used after without being checked if it can be undefined

Only if the device configuration is malformed on the firmware side.

> i'm pretty sure there is a good reason why a streamId can be null , but just mentioning it here in case it wasn't meant to be

yes, not every event can be associated with a stream!

PR #403, opened 2020-04-14
Make tests async 

### LINKED ISSUE #380: make hambda tests async
you can reference this pr comments to understand that we are losing almost a factor of 4 when using sync tests

https://github.com/ospin-web-dev/hambda/pull/370

whoever will work on this ticket should also try to add the ability to fail fast that is if a test is failing we should not continue testing the others , that will save us a good amount of time

O-Susanoo:
closes #380 

this one is also making dep install async


package.json
```diff
@@ -10,8 +10,8 @@
     "postinstall": "nps installServiceDeps ; wait && nps seedAllLibFiles",
     "test": "jest ./test/* -i --forceExit",
     "testWithCoverage": "npm test -- --coverage",
-    "testAll": "npm test && nps testServices",
-    "testAllWithCoverage": "npm run testWithCoverage && nps testServicesWithCoverage",
+    "testAll": "nps testAll",
+    "testAllWithCoverage": "nps testAllWithCoverage",
```
DanielSeehausen, line 14: <3

O-Susanoo:
already did if you see the commit history you will see that i made a failing one just to make sure that it's being caught

PR #404, opened 2020-04-15
287 remove dashboard streams from device

vdiezel:
_My necessaries are embark'd: farewell._
                                     - dashboardStreams

cleaned up the DB too.
Since I'm the only one creating devices, no new devices with the `dashboardStreams` will be created even before this will be merged (I'm using dev anyway for device creation)


vdiezel:
replaced by https://github.com/ospin-web-dev/hambda/pull/405

PR #405, opened 2020-04-15
287 remove dashbaord streams

### LINKED ISSUE #287: remove dashboardStreams from device collection
now that the dashboard configuration is user specific we should remove old device data configuration from mongodb

(ensure code on reactor and hambda is rdy for the migration)

vdiezel:
_My necessaries are embark'd: farewell._
                                     - dashboardStreams

cleaned up the DB too.
Since I'm the only one creating devices, no new devices with the `dashboardStreams` will be created even before this will be merged (I'm using dev anyway for device creation)
closes #287


O-Susanoo:
I hope that this one is also deleted from all devices in db

PR #410, opened 2020-04-22
379 abstract user model sausage hidden

### LINKED ISSUE #376: Move User authorization calls out of Model functions
related to #377 
related to #378 
blocked by #379 

Right now a lot of our model methods are doing user authorization directly in them.
(Note: this does not include dependency assertion, e.g.: 'a device needs to be found by that id for you to perform this authorization on a process`)

**Choose one of:**
**OR**
[Give this a read!](https://martinfowler.com/bliki/AnemicDomainModel.html)
One path forward we could consider is to either make a new Class extend the core User interface (or use connectable on the base Class). This extension Class would have methods that look like what we already have on the User class (functionality combined with authorization). These methods would continue to take a userId and do the same stuff on the model. This would save us from 'robbing ourselves blind' as Fowler put it. 

### LINKED ISSUE #379: Transition `user` to a `User` model
Continue that transition to a more functional interface that matches the other models (Event, UIConfig, etc. should be decent examples). the user/src/user/user.js file should be removed after this ticket is finished
- User should become a connectable as part of this transition
- Update the seeder to make use of User.schema instead of making User through a different route
- click test thoroughly! add any high value tests you see missing!
- Update all lambda functions only to call the new model (breaking out authorization into the controllers will be in another ticket)

Note: this is a huge ticket that touches a lot, but it is not a great candidate to get merged at different times. Understanding this, consider making a detailed commit plan before starting.

### LINKED ISSUE #384: Connect all models to DB when we connect to the DB
### Justification

We'd like to continue moving away from interfacing with data directly through calls to mongodb (instead of calls to the models, which standardize their call to mongodb). However we do this, it needs to be standard, abstracted, and easy for developers. Details re: connecting to the DB, model's connecting to the DB, etc. should be abstracted. 

Now that models are being tightened up to provide shareable interfaces to their data, we have some great opportunities for easy improvements.

The goal should be, if one day we wanted to move to something like mongoose or other ORM/ODM, it would be relatively easy to do, because our code already has a comprehensive Database-Model-Controller separation. This ticket is another step in moving in that direction. 

### Background

Currently, we have some boiler plate code in the form of:
- Lambda function starts 
- we connect to DB (which is blocking)
- **we connect some models to that DB. this will increase to be 2 - 3 models in a lot of controllers** <-- this part is instant and not time consuming

Where the last step is the opportunity to reduce boilerplate code, and can be combined with step the connection to the DB

### Deliverables

- [ ] Add a new lib file + function(s) that connects all model's to a database (the function can be passed the database)
- [ ] require the new file in the mongo-db lib, and add a method `connectDB`?, which does the normal `getDB` call + the new module added in the step above
- [ ] update controllers (lambda function entry points) to await this instead of `getDB` (only lambda functions where a model is being `connect`ed need to be replaced for the time being, the rest will be refactored in time. `services/event/src/event/delete.js` is an example of one that would be replaced)

DanielSeehausen:
closes #379 
closes #376 
closes #384 

I apologize for the size! There was a lot that was coupled and needed to be done to make this change. I popped everything out of staging and added selectively in a (hopefully) easier order to follow.

Reading through commit wise is a must if you do choose to read. Please see commits for more specific messages!
if you don't want to read through the whole thing, (which is totally understandable), please just leave a note what commits you were able to look at.

I will be click testing a lot of this tomorrow.


lib/ResourceAccess/Access.js
```diff
@@ -0,0 +1,70 @@
...
+    if (!group) return false
+
+    return Group.userPresent(group, userId)
+  }
+
+  static addUserToGroup(access, userId, groupName) {
+    return access.map(group => (
+      group.name === groupName
+        ? Group.addUser(group, userId)
+        : { ...group }
```
vdiezel, line 49: I think in functional programming it is okay to not copy the array elements that remain untouched - at least this was the pattern I have learned: 
https://redux.js.org/recipes/structuring-reducers/immutable-update-patterns

see section updating an item in an array: 
```js
function updateObjectInArray(array, action) {
  return array.map((item, index) => {
    if (index !== action.index) {
      // This isn't the item we care about - keep it as-is
      return item
    }

    // Otherwise, this is the one we want - return an updated value
    return {
      ...item,
      ...action.item
    }
  })
}
```

DanielSeehausen:
just updated codeclimate to watch staging

lib/ResourceAccess/Group.js
```diff
@@ -0,0 +1,44 @@
+const Joi = require('@hapi/joi')
+
+class Group {
+
+  // static get GROUP_TYPES() { [ 'operator viewer admin w/e' ] }
+  static get SCHEMA() {
+    return Joi.object({
+      name: Joi.string().min(1).max(45).regex(/^\S+$/).required(),
+      users: Joi.array().items(Joi.string(), Joi.number()).default([]),
```
vdiezel, line 9: why do we allow `Joi.number` here? userIds are always uuidv4

lib/ResourceAccess/Group.js
```diff
@@ -0,0 +1,44 @@
...
+  static create(name, rights) {
+    return Group.schema({ name, rights, users: [] })
+  }
+
+  static userPresent(group, userId) {
+    return group.users.includes(userId)
+  }
+
+  static addUser(group, userId) {
+    if (Group.userPresent(group, userId)) return { ...group }
```
vdiezel, line 27: also here we don't need to copy the group in the case the user is already present, just return as is (see comment in `Access` model.

.eslintrc.js
```diff
@@ -17,10 +18,7 @@ module.exports = {
       "named": "ignore",
       "asyncArrow": "always"
     }],
-    "no-underscore-dangle": [
-      "error",
-      { "allow": ["__MONGO_URI__", "__MONGO_DB_NAME__"] }
-    ],
+    "no-underscore-dangle": "off",
```
O-Susanoo, line 21: Like this one :+1:

test/lib/ResourceAccess/Access.test.js
```diff
@@ -0,0 +1,146 @@
...
+      access = Access.addUserToGroup(access, userId, 'admins')
+      access = Access.addUserToGroup(access, userId, 'operators')
+      expect(Access.userIsInGroup(access, userId, 'admins')).toBe(true)
+      expect(Access.userIsInGroup(access, userId, 'operators')).toBe(true)
+
+      access = Access.removeUserFromGroup(access, userId, 'admins')
+      expect(Access.userIsInGroup(access, userId, 'admins')).toBe(false)
+
+      access = Access.removeUserFromGroup(access, userId, 'operators')
+      expect(Access.userIsInGroup(access, userId, 'admins')).toBe(false)
```
vdiezel, line 127: here you want to check if a user is in operators I assume :)

test/lib/ResourceAccess/Access.test.js
```diff
@@ -0,0 +1,146 @@
...
+
+      expect(accessUsers).toEqual(expect.arrayContaining([
+        adminIdA, adminIdB, operatorId, viewerId,
+      ]))
+    })
+  })
+
+
+  describe('getUsersGroups', () => {
+    it('returns an array of the access groups a user in in', () => {
```
vdiezel, line 43: `in in`

DanielSeehausen:
hopefully just updated codecov to use staging as well

services/user/src/user/acceptDeviceInvitation.js
```diff
@@ -1,26 +1,74 @@
...
+  const device = await Device.find(deviceId)
+  assertMany([
+    () => canFind(device, `Device: ${deviceId}`),
+    () => resourceHasAccessGroup(device, groupName),
+  ])
+
+  // TODO: make whole thing a task
+  const deviceUsers = DeviceAccess.getUsers(device.access)
+
+  try {
```
ConorTighe1995, line 59: Candidate for a function?

lib/ResourceAccess/Access.js
```diff
@@ -0,0 +1,70 @@
...
+  static schema(groups) {
+    return Joi.attempt(groups, Access.SCHEMA)
+  }
+
+
+  static getGroupNames(access) {
+    return access.map(group => group.name)
+  }
+
+  static getUsersGroups(access, userId) {
```
vdiezel, line 19: With our current access rights definition it doesn't make sense to have a user in multiple groups, because the access rights are somewhat in a hierarchy. When we want to have users to be present on multiple groups, we need another access rights concepts that builds on composition (and `operator` would get only operating rights, but for viewing processes they would also need to be on the `viewer` group while now the `operator` group already includes view rights. 

We can keep the code as it is, but until we haven't finally designed authorization we should never have a user be present on multiple groups.

lib/access-rights-lib.js
```diff
@@ -196,18 +130,8 @@ function processAuthorizer(token){
...
-    [RIGHTS.EXECUTE]: token[2] === "1",
-    [RIGHTS.DELETE]: token[3] === "1",
-    [RIGHTS.MANAGE_ACCESS]: token[4] === "1"
-  }
-}
-
-attachAuthorizer(RESOURCE.PROCESS, processAuthorizer);
-attachAuthorizer(RESOURCE.DEVICE, deviceAuthorizer);
+attachAuthorizer(RESOURCE.PROCESS, resourceAuthorizer)
+attachAuthorizer(RESOURCE.DEVICE, resourceAuthorizer)
```
vdiezel, line 134: :gun:

services/user/src/user/create.js
```diff
@@ -19,22 +20,21 @@ async function create(event, context) {
...
 
-  const result = await user.create(newUserData)
+  await connectAllModels()
+
+  const newUser = await User.create(newUserData)
 
   context.succeed(event) // triggers lambda to close via cognito request method
 
-  return success(result)
+  return success(newUser) // this return should not be doing anything if context.succeed is working as expected
```
ConorTighe1995, line 37: If context fails or errors will this still return as usual? I had a quick look around for http-response in my IDE and the package.json of this repo and couldn't see it, does it have a failed method? would a flow like this possible:

```
try{
  context.succeed(event)
  return success(newUser)
}catch(e){
  return failed(e)
}
```

lib/models/User/Notifications/Notification.js
```diff
@@ -0,0 +1,45 @@
...
+
+
+class Notification {
+
+  static get NOTIFICATION_TYPES() {
+    return [ 'persistent', 'ephemeral' ]
+  }
+
+  static get NOTIFICATION_TOPICS() {
+    return [ 'device/invitation', 'changelog' ]
```
vdiezel, line 14: we also need `device/revocation` as a topic here, otherwise our device access revocations will not work!

lib/models/User/Notifications/Notification.js
```diff
@@ -0,0 +1,45 @@
...
+  // TODO: sub schemas needed for revocation, invitation, changelog notifications
+  static get SCHEMA() {
+    return Joi.object({
+      id: Joi.string().pattern(UUIDV4_REGEX),
+      topic: Joi.string().valid(...Notification.NOTIFICATION_TOPICS).required(),
+      type: Joi.string().valid(...Notification.NOTIFICATION_TYPES).required(),
+      from: Joi.object({
+        type: Joi.string().required(),
+        name: Joi.string().required(),
+        originId: [ Joi.number(), Joi.string() ],
```
vdiezel, line 26: we miss the keys `sourceName` and `sourceType` here - they are being destructured here: 
https://github.com/ospin-web-dev/hambda/blob/2ac7007c46e96f19398727df3ccc2b4ee9f0f49d/services/user/src/user/patchAllNotifications.js#L9

but actually never used. The vales they curently have are those: 
https://github.com/ospin-web-dev/REACTor/blob/master/.circleci/publishChanges.js#L35

lib/models/User/Notifications/Notification.js
```diff
@@ -0,0 +1,45 @@
...
+      }).required(),
+      data: Joi.object({
+        groupName: Joi.string(), // Only on invitations TODO: this needs to be made a sub schema when device model is compelted
+        deviceId: Joi.string().pattern(UUIDV4_REGEX), // Only on invitations TODO: this needs to be made a sub schema when device model is compelted
+        message: Joi.string().default(''),
+      }).required(),
+      read: Joi.bool(), // TODO: is this still required?
```
vdiezel, line 33: I think so - this should be used e.g. for device invitations: they are persistent, but as long as they are unread, the notifications icon on the side bar in REACTor remains red. Once read, the icon turns grey. Did this to prevent annoying users when they have persistent messages.

lib/ResourceAccess/Group.js
DanielSeehausen: that's a good question
I was picturing access as a general thing, where the `users` just needs to be unique identifiers. If it were used with another resource that doesn't happen to use uuidv4, I wanted it to be compatible. I was trying to picture if this were an npm package, what it would need to be usable by anything with a `.access` on it

.eslintrc.js
DanielSeehausen: that one goes out to you, @O-Susanoo

services/user/src/user/getChangelog.js
```diff
@@ -1,14 +1,19 @@
...
+  const changelogs = await GithubAPI.getChangelogs('REACTor')
+  const prs = changelogs.map(serializePR)
+
   return success(prs)
 }
 
 exports.main = HTTPDefaultMiddleware(getChangelog, {
   useDoNotWaitForEmptyEventLoop: false,
-  useHTTPEventNormalizer: false, // why?
+  useHTTPEventNormalizer: false, // why? -VD || 'I dont know either' -DS
```
ConorTighe1995, line 17: Do you know the after effects of setting this to false? How will how will it normalizer the data?

lib/ResourceAccess/Access.js
DanielSeehausen: that makes sense - I will update now

lib/ResourceAccess/Group.js
DanielSeehausen: updated

test/lib/ResourceAccess/Access.test.js
DanielSeehausen: very good eye thank you
PR review WORKS

DanielSeehausen: updated

test/lib/ResourceAccess/Access.test.js
DanielSeehausen: not pedantic at all
good 👁️

services/user/src/user/revokeIotPolicy.js
```diff
@@ -1,27 +1,60 @@
...
+
+  const user = await User.find(targetUserId)
+
+  assertMany([
+    () => canFind(user, `User ${targetUserId}`),
+    () => isMe(requestingUserId, targetUserId),
+  ])
 
-  validateRevokeIotPolicy(requestData)
+  await IotAPI.detachDevicePolicyFromUser(revocation.data.deviceId, cognitoIdentityId)
```
ConorTighe1995, line 52: Maybe my eyes are playing tricks on me, but you should switch line 52 and 54 as revocation is being hoisted

lib/tasks/users/destroyUser.js
```diff
@@ -0,0 +1,40 @@
...
+
+async function removeUserFromProcesses(user) {
+  return Promise.all(user.processes.map(processId => (
+    Process.removeUserAccess(processId, user.id)
+  )))
+}
+
+
+module.exports = async userId => {
+  /* We need to do the following cleanup:
```
ConorTighe1995, line 21: Do theses comments still need to be here now that the below code is in place?

services/user/src/user/delete.js
```diff
@@ -1,22 +1,44 @@
...
+    () => canFind(targetUser, `User ${targetUserId}`),
+    () => canFind(requestingUser, `User ${requestingUserId}`),
+    or([
+      () => isMe(requestingUserId, targetUserId),
+      () => hasAdminAccess(requestingUser),
+    ]),
+  ])
 
-  const result = await user.delete(requestData)
+  // Note: this is more than removing the user from the db. See the task
```
ConorTighe1995, line 40: If this requires a note should then name be more descriptive? something like removeUserFromSystem or deleteAllUserRelatedData

services/user/test/user/delete.test.js
```diff
@@ -6,8 +6,6 @@ const Process = require('models/Process/Process')
 const Seeder = require('test-helpers/seeder/Seeder')
 const mongo = require('mongodb-lib')
 const CognitoAPI = require ('cognito/CognitoAPI')
-//process.env.AWS_COGNITO_USERPOOL_ID = 'MERKEL!'
```
ConorTighe1995: lol

services/user/src/user/acceptDeviceInvitation.js
DanielSeehausen: that's a good idea - I have a lot of controller stuff that is way longer than it should be. I updated - will go up in new commit ASAP

lib/tasks/users/destroyUser.js
```diff
@@ -0,0 +1,40 @@
...
+     // user.processes.forEach Process.removeUser(userId)
+   * 3. Delete the user from the database
+     // User.delete(id)
+     // CognitoAPI.deleteUser(userId)
+   * 4. Delete the user from the cognito userpool
+  */
+  const user = await User.find(userId)
+  await Promise.all([
+    removeUserFromDevices(user),
+    removeUserFromProcesses(user),
```
vdiezel, line 34: Nesting `Promise.all` - so easy!

lib/ResourceAccess/Access.js
DanielSeehausen: that's a good point, since ours is strictly a hierarchy. Maybe I should do one of these two things for now until we have further definition?:

These are both low effort options:
- ensure you can't add a user to a group if they are already found on the access object
- always switch the user out of group existing to group target if someone tries to add them to group target

What do you think @vdiezel ?

services/user/src/user/create.js
vdiezel: We originally had this flow you describe here in all functions. We decided: If it fails, let it fail. Otherwise you can use `context.fail` in teh catch block.

vdiezel: this handler needs some love and investigation anyway

lib/ResourceAccess/Access.js
ConorTighe1995: If the access right are inspired by linux system permissions so I can understand why it might make sense from a design perspective as its also possible to create access groups on linux for different files and folders. But your right, from a user perspective we have this hierarchy system and then in the back-end it looks to me to be evolving into a resource orientated permissions system where each user might have very different rights to the next if they ever intersect groups.

services/user/src/user/create.js
DanielSeehausen: I'm unsure. This was really me kicking the can down the road.

I don't believe that return can ever be hit (`context.succeed` ends execution as I understand), and on top of that the response isn't necessary because the caller doesn't make use of it. I just removed that line all together.

services/user/src/user/patchAllNotifications.js
```diff
@@ -1,26 +1,46 @@
...
+    body: {
+      originId,
+      sourceName,
+      sourceType,
+      topic,
+      message,
+    },
+  } = event
+
+  // I believe any user could inject the correct sourceName/Type values and fire this! This endpoint may be lacking some security
```
vdiezel, line 20: No, this endpoint is not exposed to the public via HTTP but is called via the CD suite which has an AWS profile sufficient enough to envoke this using the `aws-sdk`. You can see here, is has no triggers: 
https://github.com/ospin-web-dev/hambda/blob/staging/services/user/serverless.yml#L368. So we are safe here and this comment can be removed :)

services/user/src/user/create.js
DanielSeehausen: didn't see viktor's comment before I wrote
like he says. I removed the misleading return for now

lib/ResourceAccess/Access.js
ConorTighe1995: Will that happen here though? if not should this be `getUsersGroup`

DanielSeehausen: so the second bullet point probably?

services/user/src/user/revokeIotPolicy.js
vdiezel: great catch - that is my bad, I made Daniel do this (because logically it makes sense). We need the `deviceId` which is stored on the revocation in the DB. Guess we have to fetch it first between lines 50 and 52.

vdiezel: I also see that the revocation notifications is actually never removed from the user here, or?

lib/ResourceAccess/Group.js
vdiezel: YAGNI it :stuck_out_tongue:

lib/models/User/User.js
```diff
@@ -0,0 +1,108 @@
+const uuidv4 = require('uuid').v4
+const Joi = require('@hapi/joi')
+
+const mixin = require('functional-helpers/mixin')
+const Connectable = require('models/Connectable')
+const UserNotifications = require('models/User/Notifications/UserNotifications')
+
+const UUIDV4_REGEX = /[a-f0-9]{8}(-[a-f0-9]{4}){3}-[a-f0-9]{12}/
```
ConorTighe1995, line 8: This appears in a few places, might be worth introducing a regex utility file where we can update this easily in the future and expand on regex functions. would DRY out the code and save time when validating and testing regex.

lib/ResourceAccess/Access.js
ConorTighe1995: @DanielSeehausen I didn't refresh my page so only seen your comment after I wrote that, yea I think that could reduce complexity and confusion in the future.

lib/models/User/Notifications/Notification.js
DanielSeehausen: added it!

lib/models/User/Notifications/Notification.js
DanielSeehausen: as I understand it, those keys are serialized in the controller:
```
 // patchAllNotifications.js
  const notification = Notification.new({
    topic,
    type: 'persistent',
    from: {
      type: sourceType,
      name: sourceName,
      originId,
    },
    data: { message },
  })
```

DanielSeehausen: and in the DB:
image

lib/models/User/Notifications/Notification.js
DanielSeehausen: 👍 noice

services/user/src/user/getChangelog.js
DanielSeehausen: the event normalizer just serializes the event for value access convenience, IIRC

ConorTighe1995: Ok just curious because of the comments

services/user/src/user/revokeIotPolicy.js
DanielSeehausen: very good catch - even the linter was complaining and I entirely did not see it.
I just reversed them. 

it shouldn't hoist in that scenario, so the code was definitely broken. At the top of the controller I mentioned this needs a 200 test...I guess the fact that such a big issue wasn't caught by a test means I should add that now.

lib/models/User/Notifications/Notification.js
DanielSeehausen: tests also seem to be using type/topic, so this is strange in general...might need to look into this more

lib/ResourceAccess/Group.js
DanielSeehausen: YAGNI is always a good point

I want to push back on this because that would be making the module match its current use cases, instead of the module being true to what it offers in general. In this case, it offers to hold a collection of unique identifiers for users. Its pretty often those are numbers. While in this case we know all the ways it is being used (only with strings, for our purposes), it's not a guarantee all of its consumers use strings. I believe making it string would make it more brittle.

lib/ResourceAccess/Access.js
DanielSeehausen: ill update it
meant to mention in the comments - would really like to make this access a key: value where:
```js
{ 
  viewers: { },
  admins: { },
}
```

at some point in the future. Hoping this is a step to standardize the interface to make that easier if it ever becomes worthwhile to do

services/user/src/user/revokeIotPolicy.js
DanielSeehausen: added tests

lib/tasks/users/destroyUser.js
DanielSeehausen: no! I removed them

services/user/src/user/delete.js
DanielSeehausen: good idea - i will go for `deleteAllUserRelatedData`

DanielSeehausen: updated

services/user/test/user/delete.test.js
DanielSeehausen: :))

lib/tasks/users/destroyUser.js
DanielSeehausen: glad you like it

services/user/src/user/patchAllNotifications.js
DanielSeehausen: nice - thanks for the education. removing comment

lib/ResourceAccess/Group.js
DanielSeehausen: I should add I really do intend to make this an npm module

lib/models/User/User.js
DanielSeehausen: yes I agree - have been lazy for quite some time just always copying this.

While tempting to monkey patch String, I'll instead add it to a util or something :)

DanielSeehausen:
thank you guys very much for fast review - I updated and made all change requests except for one (re: access unique identifiers - I left a comment)

The biggest changes are to Access. it is now implemented to only allow a single user. See tests for documentation on that

services/user/src/user/revokeIotPolicy.js
ConorTighe1995: Ahh yes hoisting is var only, good point

lib/models/User/User.js
DanielSeehausen: i added this!

PR #420, opened 2020-04-30
Jsdocs for events

### LINKED ISSUE #412: JSDocs for all event triggers
We need to provide an API contract for/between firmware for all device <-> hambda communication

JSDocs seems like a good option.

We need a solution that, in the CI/CD pipeline, generates the docs file somewhere that can easily be seen by other members of the webdev + firmware team.

- [x] add jsdoc dependency
- [x] Investigate and choose an easy and automated option possible for hosting the jsdocs where others can see them
- [x] implement auto deploy/host/update of these docs as part of the CI/CD pipelines
- [ ] update events documentation with daniel for firmware

vdiezel:
- Implements JSDocs into hambda
- Adds documentation for all mqtt messages
- Adds auto generating and deploying the docs to 
http://docs.ospin-services.com/mqtt.html

I does look very nice and professional, don't know whats your take here on the comments within the code, they are cluttering the code a bit, but I think I can tolerate that...

closes #412 

Once this got merged you will also need to update your .env files with the bucket name that hosts the docs. Not really important but might be helpful when you want to deploy from your local machine.  I added it already to the pinned .env files.


package.json
```diff
@@ -35,14 +36,16 @@
   },
   "devDependencies": {
     "@shelf/jest-mongodb": "^1.1.5",
+    "concurrently": "^5.1.0",
```
vdiezel, line 39: no clue why it moved those around..

package-scripts.js
```diff
@@ -17,6 +17,13 @@ const SERVICE_NODE_MODULE_PATHS = (
     .join(' ')
 )
 
+const SERVICE_SRC_PATHS = (
+  SERVICES
+    .filter(serviceName => serviceName !== 'cortex')
```
vdiezel, line 22: I excluded cortex because it does not follow the same schema yet for project directories. I will take care of this later (I'm planning to move all code from Cortex back to lambda anyway...)

lib/IotDataAPI.js
```diff
@@ -14,6 +14,20 @@ const RETURN_DISPATCH = {
...
 }
 
+/**
+ * @namespace mqtt
+ */
+
+/**
+ * Device heartbeat to communicate if the device is connected to the IoT Broker;
+ * expected payload is an empty object!</br>
+ * <i>For simplicity we document the expected JSON after parsing the message. </i>
```
DanielSeehausen, line 24: you can probably remove the JSON statement from all of these!

package-scripts.js
DanielSeehausen: 👍

services/device/src/device/triggers/createEvent.js
```diff
@@ -8,6 +8,38 @@ const validateCreateEvent = require('../../joi/triggers').createEvent
 const Device = require('../device')
 const Logger = require('../../logger/Logger')
 
+/**
+ * Creates a device event</br>
```
ConorTighe1995, line 12: Is there a missing <b> here

services/device/src/command/command.js
```diff
@@ -49,6 +90,20 @@ class Command {
     return await IotDataAPI.publishCommand(deviceId, payload)
   }
 
+  /**
+   * Communicates a 'pause process' command to a device</br>
```
ConorTighe1995, line 100: I could be wrong but it looks like a <b> is missing from each one as a copy and paste error.

services/device/src/device/triggers/createEvent.js
vdiezel: it should be <br> instead if </br> I guess - looks like it didn't have any effect. I'll check it!

vdiezel: removed them!

.circleci/config.yml
```diff
@@ -147,3 +171,7 @@ workflows:
           filters:
             branches:
               only: master
+      - deployDocs:
+          context: hambda-deployment-prod
+          requires:
+            - deploy-prod
```
O-Susanoo, line 177: should we also trigger in staging maybe other team may be interested in beta functionality ?

package.json
O-Susanoo: great to have npm in our back

services/device/src/command/command.js
```diff
@@ -30,12 +57,22 @@ class Command {
...
   }
 
+  /**
+   * Communicates a 'end process' command to a device
+   * <h5>Topic</h5> <code style='color: #c7254e;'>{env}/command/{deviceId}</code>
+   * @name Command: End Process
+   * @memberOf mqtt
+   * @type {object}
+   * @property {string} command - valid values: end
+   * @property {string} processId - associated process ID (uuidv4)
```
O-Susanoo, line 73: Nice very intuitive

services/device/src/command/command.js
```diff
@@ -113,6 +183,26 @@ class Command {
     return await IotDataAPI.publishCommand(deviceId, payload)
   }
 
+  /**
+   * Communicates a 'open SSH tunnel' command to a device
+   * <h5>Topic</h5> <code style='color: #c7254e;'>{env}/command/{deviceId}</code>
```
O-Susanoo, line 188: like how we can inject css code

lib/IotDataAPI.js
```diff
@@ -69,13 +83,37 @@ class IotDataAPI {
     return this.execute(mqttMsg)
   }
 
+  /**
+   * Communicates user defined device settings to the device after device bootup</br>
+   * <i>For simplicity we document the expected JSON after parsing the message. </i>
+   * <h5>Topic</h5> <code style='color: #c7254e;'>{env}/deviceDescription/settings/{deviceId}</code>
+   * @name Update Device Description Settings
+   * @memberOf mqtt
+   * @deprecated possible not necessary as it currently communicates only calibrations which
```
ConorTighe1995, line 89: this looks like it should be 'possibly'

ConorTighe1995: is this indicating to the reader that `publishStoredDeviceConfiguration` is deprecate or will be?

.circleci/config.yml
vdiezel: I agree but we probably need to different spots to show the documentation once this is used publicly!

lib/IotDataAPI.js
vdiezel: I'm just abusing the deprecated label here as a reminder. It still not decided whether we will deprecate this message or not...
And you are correct on the possibly!

vdiezel:
> This is great and documentation is always a pain to write so I appreciate the detail and effort put in. I left some comment on small things.
> 
> As this matures should we have a developer section on the Ospin sidebar to point developers to our public APIs and documentation? or is this strictly internal?

For now it will be only used internally. Later we will need to come up with somethin smarter, like model docs for us, mqtt for firmware engineers, HTTP API fore web devs, and then also different documentation versions...

PR #422, opened 2020-05-01
390 device admins can run all commands

### LINKED ISSUE #390: Device admins should be able to run commands on any device process
### Description
Currenttly, the device admins can't run commands (start, pause, resume, end) on device processes where they are not present on (because created by another device user). But they should!

---

### Proposed Solution(s)
In the same manner we allow device admins access to all processes on the process service, the commands in the device service should allow running any command on a device process when you are a device admin. 

---

### Deliverable(s)
- affected commands: start, pause, resume, end, update (same function as start), calibrate

---

vdiezel:
closes #390 

Changes access control to commands to allow device admins to execute commands on device processes they have no regular access to. 

Adds test coverage to commands
Adds JOI validartion to commands
Shows how to compose Joi schemas for MAXIMUM DRYNESS (last commit). 

Line count due to tests!


services/device/src/joi/command.js
```diff
@@ -0,0 +1,14 @@
+const Joi = require('@hapi/joi')
+
+const { createSchemaAssertion } = require('validation-lib')
+const { UUIDV4_REGEX } = require('utils')
```
DanielSeehausen, line 4: you found it!
man thought I'd get someone on this one...

services/device/test/command/end.test.js
```diff
@@ -0,0 +1,196 @@
...
+
+const endProcess = promisifyHandler(require('../../src/command/endProcess').main)
+const { endProcessEvent } = require('./seeds/event-generators/endProcessEvent')
+
+
+describe('Command', () => {
+
+  beforeAll(async() => {
+    const db = await mongo.getDB()
+    Seeder.connect(db)
```
DanielSeehausen: see `connectAllModels`
the db and seeder calls can be replaced with that here. It will help avoid accidentally missing connecting models going forward too

services/device/test/command/end.test.js
```diff
@@ -0,0 +1,196 @@
...
+      const res = await endProcess(event, {})
+
+      return { res, body: tryParseJSON(res.body) }
+    }
+
+    describe('returns 200 and publishes the end command', () => {
+      test('when the user is a device admin but not present on the process', async () => {
+        const user = await Seeder.seedUser()
+        const deviceAccess = Access.addUserToGroup(DeviceAccess.create(), user.id, 'admins')
+        const device = await Seeder.seedUserDevice(user, { deviceData: { access: deviceAccess } })
```
DanielSeehausen: See `Device.addUserTo`

services/device/test/command/end.test.js
```diff
@@ -0,0 +1,196 @@
...
+        expect(payload).toEqual({ command: 'end', processId: process.id })
+      })
+
+      test('when the user is not a device admin, but has sufficient resource access on device and process', async () => {
+        const user = await Seeder.seedUser()
+        const processAccess = Access.addUserToGroup(ProcessAccess.create(), user.id, 'admins')
+        const deviceAccess = Access.addUserToGroup(DeviceAccess.create(), user.id, 'operators')
+        const device = await Seeder.seedUserDevice(user, { deviceData: { access: deviceAccess } })
+        const process = await Seeder
+          .seedProcess(user, device, { processData: { access: processAccess } })
```
DanielSeehausen: See `Device.addUserTo`
consider adding also an `addUserTo` to the process model (will conflict with the Process overhaul PR but ill happily deal with that when it comes time)

services/device/test/command/end.test.js
```diff
@@ -0,0 +1,196 @@
...
+    })
+
+    describe('returns 403 and does not publish the end command', () => {
+      test('when the user has sufficient device access (but not an admin) but insufficient process access', async () => {
+        const user = await Seeder.seedUser()
+        const processAccess = Access.addUserToGroup(ProcessAccess.create(), user.id, 'viewers')
+        const deviceAccess = Access.addUserToGroup(DeviceAccess.create(), user.id, 'operators')
+        const device = await Seeder.seedUserDevice(user, { deviceData: { access: deviceAccess } })
+        const process = await Seeder
+          .seedProcess(undefined, device, { processData: { access: processAccess } })
```
DanielSeehausen: Same thing as above

services/device/test/command/end.test.js
```diff
@@ -0,0 +1,196 @@
...
+        const { res } = await fireEndProcessEvent(device, process, user)
+
+        expect(res.statusCode).toBe(403)
+        expect(MockIotDataAPI.publishCommand.mock.calls.length).toBe(0)
+      })
+
+      test('when the user has sufficient device access (but not an admin) but no process access', async () => {
+        const user = await Seeder.seedUser()
+        const deviceAccess = Access.addUserToGroup(DeviceAccess.create(), user.id, 'operators')
+        const device = await Seeder.seedUserDevice(user, { deviceData: { access: deviceAccess } })
```
DanielSeehausen: same thing as above - I won't keep writing it for every example

services/device/src/joi/command.js
```diff
@@ -3,44 +3,26 @@ const Joi = require('@hapi/joi')
...
   entryPhaseId: Joi.number().required(),
   elapsedTime: Joi.number().required(),
-}))
+})))
 
 module.exports = {
-  endProcess,
+  endProcess: createSchemaAssertion(commonSchema),
+  resumeProcess: createSchemaAssertion(commonSchema),
+  nextPhase: createSchemaAssertion(commonSchema),
```
DanielSeehausen, line 25: pretty g4m3r. Nice!

services/device/src/command/command.js
```diff
@@ -35,10 +41,10 @@ class Command {
     const process = await this.processesDB.findOne({ id: processId })
```
DanielSeehausen, line 41: see `Process.find`

services/device/src/command/command.js
```diff
@@ -93,10 +99,10 @@ class Command {
...
```
DanielSeehausen, line 99: see `Process.find`

services/device/src/command/command.js
DanielSeehausen: we are in a tricky place with this. This pr is good and focused and I don't want to distract it with updating the usage of model stuff. 

That being said, we really do need to update this to use the interfaces soon.
I am going to move this review back to approved and am just requesting that you make a ticket similar to the process/device/user model tickets to move validations to the controllers and to delegate to the model/interface methods for data access

DanielSeehausen: (that being said, the tests should be good to be updated with the new stuff where applicable. see inline comments)

services/device/test/command/end.test.js
vdiezel: But this way we actually might miss a missing connection to the DB in the source code because tests do not fail because the models are connected in the test file. So I feel better not connecting the models here but delegate that to the source code instead.

vdiezel:
> This looks great.
> 
> The additional test coverage is awesome and overdue. Feels very good to get this in.
> Would this stuff also warrant updating JSdocs depending on what goes in first?
> 
> the only requests I have is to update the test files with few minor things mentioned inline

This doesnt require changes to the JSdocs. 
Because this was part of the device service, I didn't bother with new models yet beyond Access

services/device/src/command/command.js
vdiezel: command is part of the device service I though it will be covered when you are going to touch the Device model. I will do that.

services/device/test/command/end.test.js
vdiezel: I think this should be done once the model is implemented in the commands because it would require me to connect the models to the db in the test file and for the same reason as described in the comment above I'd like to prevent this!

PR #423, opened 2020-05-03
Kill cortex

vdiezel:
I have killed cortex on elastic beanstalk. And from the ashes a new lambda service rises by the name of 'data'.

The reason was that the downsides of it were bigger than the advantages. The main reason to use a server instance was a.) to have a redis instance running on it,  b.) having some worker running on it to clean up cache if necessary and c.) reducing amount of lambda invocations. But redis is now running  remotely and our concerns about performance weren't really justified: Lambda functions accessing redis for the event triggers e.g. need only 2-3ms for execution, so about 7-10 times faster than accessing mongoDB.
For the worker to clean up the cache from stale data I have discovered that lambda provides you the possibility running lambda functions on schedule: See here: 

https://github.com/ospin-web-dev/hambda/blob/c92e94bdd019fddd939b88a46626cc0e8c029a64/services/data/serverless.yml#L101

ziemlich geil, oder?

Also with a seperate server instance for every environment some were running unused for quite a while, generating cost (especially logs) - with lambda we will only pay what we use and it turned out that lambda invocations are incredibly cheap - our main cost drivers so far are CloudWatch and IoT Core.

I migrated the code to the new service which was surprisingly easy due to the connectable mixin and test coverage!

So far it still just stores data into a `testData` collection. I would like to migrate this is the next sprint.

---

### Update
renames the service to dataPoints


DanielSeehausen:
> And From the ashes you shall have a 💯 point
> 
> Dumbledore_escape

strong gif

DanielSeehausen:
image

services/data/serverless.yml
```diff
@@ -0,0 +1,111 @@
...
+      DB_USER: ${{env:DB_USER}}
+      DB_PASSWORD: ${{env:DB_PASSWORD}}
+      DB_NAME: ${{env:DB_NAME}}
+      REDIS_PORT: ${{env:REDIS_PORT}}
+      REDIS_URL: ${{env:REDIS_URL}}
+  evacuateStaleData:
+    role: AWSLambdaLoggingRole
+    handler: src/data/jobs/evacuateStaleData.main
+    events:
+      - schedule: rate(1 hour)
```
DanielSeehausen, line 105: very cool

services/data/test/data/trigger/evacuateLastChunk.test.js
```diff
@@ -0,0 +1,77 @@
...
+    data: { transition: 'finished' },
+  })
+
+  describe('evacuateFinishedProcess', () => {
+    test('evacuates the data of a finished process and removes it from the cache', async () => {
+      const streamData = getStreamData()
+      const { activeProcessIds, activeProcessesData } = getActiveProcesses(streamData)
+
+      activeProcessIds.forEach((processId, idx) => {
+        redis.set(processId, JSON.stringify(activeProcessesData[idx]))
```
DanielSeehausen, line 58: If possible, avoid going directly to the databases, and default to going through the interface/model methods

PR #426, opened 2020-05-05
387 change event revocation

### LINKED ISSUE #387: Change events so firmware can close events!
### Description

Denis/firmware requested some changes to the way events are closed.
Currently, events are deleted when they have a name that automatically resolves other events (see `EventRevoker` and how it is used through the device trigger `createEvent`)

They would like to be able to close events via an additional key on the createEvent payload.

---

### Deliverable(s)
- [ ] write tests to ensure events created with `revokeEvents: [<eventName>, <eventName>]` properly revokes events
- [ ] update event payloads with a `revokeEvents` key:
```js
// something like:
revokeEvents: Joi.array().items(Joi.string()).default([]),
```
- [ ] update endpoints to validate, accept, and handle this accurately
- [ ] remove the event recocation map and its usage throughout the code base entierly, as it is no longer required

Note: some older events won't have this key, so we need safe navigation wherever `revokeEvents` is being used

vdiezel:
Delegates event revocation to the device via a new optional key 'revokeEvents'. 
I'd like to merge the JSdocs thing before this so I can update the docs!


lib/models/Event/EventRevoker.js
```diff
@@ -1,47 +1,21 @@
...
 
-  static async revokeMatchingEvents({ revokingName, deviceId }) {
-    const revocableEvents = await EventRevoker.getRevocableEvents({ revokingName, deviceId })
+  static async revokeMatchingEvents({ revokeEvents, deviceId }) {
+    const revocableEvents = await EventRevoker.getRevocableEvents({ revokeEvents, deviceId })
     const revocableEventIds = revocableEvents.map(event => event.id)
 
-    await EventRevoker.eventsDB.deleteMany({
-      id: { $in: revocableEventIds },
-      deviceId,
```
vdiezel, line 42: event IDs are unqiue, no need to query by deviceId on top

services/device/src/device/triggers/createEvent.js
```diff
@@ -43,11 +44,10 @@ async function createEvent(requestData) {
   Event.connect(db)
   EventRevoker.connect(db)
 
-  const { deviceId, processId, name } = eventData
+  const { deviceId, processId, revokeEvents } = eventData
 
-  const eventRevokesEvents = EventRevoker.isRevokingEventName(eventData.name)
-  if (eventRevokesEvents) {
-    await revokeDeviceEvents({ revokingName: name, deviceId, processId })
+  if (revokeEvents && revokeEvents.length) {
```
vdiezel, line 49: I think we have a flaw in our validation flow. `validateCreateEvent` validates against the event schema and even though though it says ` revokeEvents: revokeEvents: Joi.array().items(Joi.string()).default([])` it does not actually apply the defaults here so I had to put the safeguard in line 49.

 And it makes sense, because when you validate data you probably don't want to mutate it, that violates the seperation of concern I guess. The defaults are only applied when the schema is used to actually create an object like we do within `Event.create()`. 

So the correct way would be to keep schema for request data validation and models seperate. Then we could set default values on the request data like this: 
```js 
const eventData = {
    deviceId: requestData.deviceId,
    eventType: requestData.eventType,
    name: requestData.name,
    createdAt: requestData.createdAt,
    processId: requestData.processId,
    displayMsg: requestData.displayMsg,
    streamId: requestData.streamId,
    revokeEvents: requestData.revokeEvents,
  }

  const requestData = validateCreateEvent(eventData) // would set defaults
  // use requestData further on...
```
instead of 
```js
 const eventData = {
    deviceId: requestData.deviceId,
    eventType: requestData.eventType,
    name: requestData.name,
    createdAt: requestData.createdAt,
    processId: requestData.processId,
    displayMsg: requestData.displayMsg,
    streamId: requestData.streamId,
    revokeEvents: requestData.revokeEvents,
  }

  validateCreateEvent(eventData) //does NOT set defaults
```

lib/models/Event/EventRevoker.js
O-Susanoo: Noice

services/device/src/device/triggers/createEvent.js
O-Susanoo: what happen if you did this

```
Joi.array().items(Joi.string()).default([]).required
```
in Joi schema i think it could have saved you the safeguard you made ? but not really sure about that

vdiezel: I can't make it required for backwards compatibility - if its required I wouldn't need a default in the first place ;)

lib/models/Event/EventRevoker.js
```diff
@@ -1,47 +1,21 @@
...
     return EventRevoker.eventsDB.find({
-      name: { $in: Array.from(revocableNames) },
+      name: { $in: revokeEvents },
       deviceId,
     }).toArray()
   }
 
-  static async revokeMatchingEvents({ revokingName, deviceId }) {
-    const revocableEvents = await EventRevoker.getRevocableEvents({ revokingName, deviceId })
+  static async revokeMatchingEvents({ revokeEvents, deviceId }) {
```
DanielSeehausen, line 14: maybe this deserves a rename now to `revokeManyEvents` or `revokeManyEventsById`

services/device/src/device/triggers/createEvent.js
DanielSeehausen: that sounds good re: keeping separate. I only had them together for convenience but if its not working definitely :thumbs-up: going separate ways

ConorTighe1995, line 49: Sounds good to keep them separate, as for if(revokeEvents && revokeEvents.length) do you need to check for both? can you not just do if(!revokeEvents) and return early

vdiezel: I need to check if the key exists (for backwards compatibility) and then I check if this events comminucates any event names it revokes for code economy reasons (checking the array length), because if it doesn't, don't run that branch of the code. `revokeEvents` is an array of strings containing names for events this event revokes, a `doorClose` event could have something like this `revokeEvents: [ '"doorOpen"]` . But the event doesn't know if the doorOpen event has been fired once, multiple times or not at all in teh past. The `revokedEvents` in line 14 are then actually really events that were finally revoked. E.g. if for some reason we had 3 doorOpen events, a single doorClose event would close all of them and `revokedEvents.length` would be 3. So `revokeEvents` is the API, while `revokedEvents` are real events,

lib/models/Event/EventRevoker.js
vdiezel: done!

vdiezel:
replaced by #446

PR #431, opened 2020-05-06
Fixes wrong directory name for triggers in dataPoints service

vdiezel:
Didn't break anything as not used yet.


vdiezel:
replaced by #444

PR #446, opened 2020-05-11
387 v2 change event revocation

### LINKED ISSUE #387: Change events so firmware can close events!

vdiezel:
closes #387 
replaces #426 (workflow issues with the branches)


vdiezel:
replaced by #446

PR #449, opened 2020-05-11
Adds process interface and related interfaces

### LINKED ISSUE #374: Transition `process` to a `Process` model
The `Process` model has already been started as part of cortex work

- [ ] Continue that transition to a more functional interface that matches the other models (`Event`, `UIConfig`, etc. should be decent examples). the processes/src/process/process.js file should be removed after this ticket is finished
- [ ] Process should become a `connectable` as part of this transition
- [ ] Update the seeder to make use of `Process.schema` instead of making Processes through a different route
- [ ] click test thoroughly! add any high value tests you see missing!

Update all lambda functions only to call the new model (breaking out authorization into the controllers will be in another ticket)

Note: this is a huge ticket that touches a lot, but it is not a great candidate to get merged at different times. Understanding this, consider making a detailed commit plan.

### LINKED ISSUE #377: Move Process authorization out of base model/interface methods
related to #376 
related to #378 
blocked by #374 

Right now a lot of our model methods are doing process authorization directly in them.
(Note: this does not include dependency assertion, e.g.: 'a device needs to be found by that id for you to perform this authorization on a process`)

**Choose one of:**
**OR**
[Give this a read!](https://martinfowler.com/bliki/AnemicDomainModel.html)
One path forward we could consider is to either make a new Class extend the core Process interface (or use connectable on the base Class). This extension Class would have methods that look like what we already have on the Process class (functionality combined with authorization). These methods would continue to take a userId and do the same stuff on the model. This would save us from 'robbing ourselves blind' as Fowler put it. 

DanielSeehausen:
standardizes the process interface
closes #374 
closes #377 

@ConorTighe1995 and I paired on the bulk of this

I know it is a ton to review, and I am sorry for that. 
I hope you will find the reviewing pretty straightforward though - there shouldn't be anything tricky.

Might be easiest to split this into multiple views, especially because I don't have this split up into different commits.


lib/access-rights-lib.js
```diff
@@ -150,12 +150,6 @@ exports.group = function (resource, token) {
   };
 };
 
-exports.getAllUserIdsFromResource = resource => (
-  resource.access.reduce((acc, group) => ([
-    ...acc, ...group.users,
-  ]), [])
-)
-
```
DanielSeehausen, line 158: `Access` was made to manage these types of requests

lib/joiUtils.js
```diff
@@ -0,0 +1,9 @@
+const Joi = require('@hapi/joi')
+
+exports.unixTimeDefaultNull = () => (
+  Joi.number()
+    .integer()
+    .min(0)
+    .default(null)
+    .allow(null)
+)
```
DanielSeehausen, line 9: :))))))))))))

lib/models/DataPoints/DataPoints.js
```diff
@@ -0,0 +1,16 @@
+const Connectable = require('models/Connectable')
+const mixin = require('functional-helpers/mixin')
+
+
+class DataPoints {
```
DanielSeehausen, line 5: ready to be grown going forward
we can put stuff that was naturally finding its way into `Process` world into here now

lib/models/Device/Device.js
```diff
@@ -1,3 +1,4 @@
+const DeviceProcesses = require('models/Device/Processes/DeviceProcesses')
```
DanielSeehausen, line 1: dear reviewer
you will see this a lot

My thinking was: subdomains are nice for organization in general. It will keep our bloat down, without reducing the convenience of calling it on what semantically makes sense. I tried to use a general rule that, if I could justify putting it not in the main interface (i.e. in `DeviceProcesses` and now `Device`) then I would.

DanielSeehausen: (note: this is just what was required for `Device`, that interface has yet to be expanded)

lib/models/Process/Process.js
```diff
@@ -1,62 +1,161 @@
+const { exclude, UUIDV4_REGEX } = require('utils')
```
DanielSeehausen, line 1: this file is the 🍖 of the PR!

lib/models/Process/Process.js
```diff
@@ -1,62 +1,161 @@
...
-const { UUIDV4_REGEX } = require('utils')
 const Connectable = require('models/Connectable')
 const mixin = require('functional-helpers/mixin')
-const ProcessAccess = require('ResourceAccess/ProcessAccess')
+const { unixTimeDefaultNull } = require('joiUtils')
 
 
 class Process {
 
   static get SCHEMA() {
```
DanielSeehausen, line 15: anytime we are replacing/adding/editing a resource in mongodb, we want to ensure that update goes through the SCHEMA. The easiest way is to run it through the `schema` method.

lib/models/Process/Process.js
```diff
@@ -1,62 +1,154 @@
...
+      'id',
+      'progression',
+      'pinned',
+      'startedBy',
+      'startedAt',
+      'finishedAt',
+      'createdAt',
+    ]
+
+    const creatableProcess = exclude(process, deletableKeys)
```
DanielSeehausen, line 90: this is a util that was added awhile ago
it removes the keys in the second argument from the from the object in the first argument

lib/models/Process/Process.js
```diff
@@ -1,62 +1,154 @@
...
 
-  static async find(processId) {
-    return Process.processesDB.findOne({ id: processId })
+  static async update(processId, processData) {
+    const process = await Process.find(processId)
+
+    const updatedAt = new Date().getTime()
+    const updatedProcess = { ...process, ...processData, updatedAt }
+
+    const populatedSchema = Process.schema(updatedProcess, Process.SCHEMA)
```
DanielSeehausen, line 144: because this is an update, it is pushed through the schema (or the application errors)

lib/models/Process/Process.js
```diff
@@ -69,8 +161,54 @@ class Process {
...
+  static async findWithDevice(processId) {
+    const process = await Process.processesDB.aggregate([
+      { $match: { id: processId } },
+      { $lookup: {
+        from: 'devices',
+        localField: 'deviceId',
+        foreignField: 'id',
+        as: 'devices',
+      } },
+    ]).next()
```
DanielSeehausen, line 183: next gives the first element found in the array that `aggregate` returns

lib/models/UIConfig/UIConfig.js
```diff
@@ -19,16 +19,19 @@ class UIConfig {
     })
   }
 
+  static get CONFIGS_SCHEMA() {
+    return Joi.object({
+      deviceDashboard: DeviceDashboard.SCHEMA,
+      graphs: Graphs.SCHEMA,
+      processBuilder: ProcessBuilder.SCHEMA,
+    }).required()
+  }
```
DanielSeehausen, line 28: see below usage on line 111 for why this was broken out

lib/models/UIConfig/UIConfig.js
```diff
@@ -100,7 +103,13 @@ class UIConfig {
...
 
+  static async deleteAllUserConfigs(userId) {
+    return UIConfig.uIConfigsDB.deleteMany({ userId })
+  }
+
   static async replace(userId, deviceId, configs) {
+    Joi.attempt(configs, UIConfig.CONFIGS_SCHEMA)
+
     return UIConfig.uIConfigsDB
       .updateOne({ userId, deviceId }, { $set: { configs } })
```
DanielSeehausen, line 114: when doing updates on complex data structures, it may be more defensive to ensure we go through a schema. Otherwise we can easily end up with some pretty malformed data

lib/models/User/User.js
```diff
@@ -67,6 +68,13 @@ class User {
     return user
   }
 
+  static async findBy(query) {
+    const user = await User.usersDB.findOne(query)
```
DanielSeehausen, line 72: 😎

lib/models/User/User.js
```diff
@@ -87,7 +95,6 @@ class User {
     return users[0]
   }
 
-  // TODO: this should be in UserDevice and this really need test coverage...
   static async addDevice(userId, deviceId) {
```
DanielSeehausen, line 98: this is here and not in `UserDevices` because there is a `UserDevice` that needs some editing first to avoid confusion with two very similar files being used differently

lib/models/connectAllModels.js
```diff
@@ -11,6 +12,7 @@ const CONNECTABLE_MODELS = [
   Event,
   Process,
   UIConfig,
+  DataPoints,
```
DanielSeehausen, line 15: this model is used when a process is deleted and it needs to delete all of its DPs

DanielSeehausen:
excited to see the test coverage

lib/joiUtils.js
```diff
@@ -0,0 +1,9 @@
+const Joi = require('@hapi/joi')
+
+exports.unixTimeDefaultNull = () => (
+  Joi.number()
+    .integer()
+    .min(0)
```
O-Susanoo, line 6: is this needed tough ?

services/process/src/joi/process.js
```diff
@@ -18,7 +18,17 @@ const getPublic = createSchemaAssertion(Joi.object({
 const create = createSchemaAssertion(Joi.object({
   requestingUserId: Joi.string().required().pattern(UUIDV4_REGEX),
   deviceId: Joi.string().required().pattern(UUIDV4_REGEX),
-  data: Process.SCHEMA.required(),
+  data: Joi.object({
+    name: Joi.string().min(1).max(45).required(),
+    deviceId: Joi.string().required().pattern(UUIDV4_REGEX),
+    createdBy: Joi.string().pattern(UUIDV4_REGEX),
+    description: Joi.object().required(),
```
O-Susanoo, line 25: i think description should be a string

services/process/test/process/create.test.js
```diff
@@ -151,96 +151,96 @@ describe('createProcess', () => {
...
-    })
+    })))
   })
 
   test('should throw 422: when data.comment is provided but in the wrong format', async () => {
     const badArgs = [ 123, true, [], '*'.repeat(251) ]
     const { user, process } = await Seeder.seedUserDeviceProcess()
 
-    badArgs.forEach(async badArg => {
+    await Promise.all(badArgs.map(badArg => (async () => {
```
O-Susanoo, line 239: i guess this is the tip of the iceberg we are doing this in many places in tests we should Promise All of them

lib/tasks/users/deleteAllUserRelatedData.js
```diff
@@ -16,13 +17,18 @@ async function removeUserFromProcesses(user) {
   )))
```
O-Susanoo, line 17: how come this is not throwing an error

lib/models/Device/Processes/DeviceProcesses.js
```diff
@@ -0,0 +1,12 @@
+class DeviceProcesses {
+
+  static async removeProcessFromAllDevices(processId) {
+    return this.devicesDB.updateMany(
+      {},
+      { $pull: { processes: processId } },
+    )
+  }
+
+}
```
vdiezel, line 10: a process can live only on one device so far

lib/models/Process/Process.js
```diff
@@ -1,62 +1,154 @@
...
-      name: Joi.string().min(1).max(45).required(),
+      id: Joi.string().required().pattern(UUIDV4_REGEX),
       deviceId: Joi.string().required().pattern(UUIDV4_REGEX),
       createdBy: Joi.string().pattern(UUIDV4_REGEX),
+      ownerId: Joi.string().pattern(UUIDV4_REGEX),
+      name: Joi.string().min(1).max(45).required(),
       description: Joi.object().required(),
-      deviceDescription: Joi.array().required(),
       entryPhaseId: Joi.number().strict().required(),
+      deviceName: Joi.string().strict(), // TODO: deprecated. remove once investigated. we have a ticket for this
```
vdiezel, line 25: I removed deviceName from the process model. I Ag'ed in hambda for `deviceName` and couldn't find it anywhere besides configurator where it is used in a different context. Did deviceName popped up somewhere due to a rebase issue on your code? You can get rid of it.

lib/models/UIConfig/UIConfig.js
vdiezel: I recall a young woman calling me pretty.........................................................................malformed!

services/process/src/joi/process.js
vdiezel: its the process description, so its an object (with phaseIds as the keys)

lib/models/Process/Process.js
```diff
@@ -1,62 +1,154 @@
...
+    return this.processesDB.updateOne(
+      { id: processId },
+      { $push: { progression: newPhase } },
+    )
+  }
 
-    return streamIds.reduce((acc, streamId) => {
-      const triggerTypes = Object.keys(eventTriggers[streamId])
+  static async find(processId) {
+    const process = await Process.processesDB.findOne({ id: processId })
```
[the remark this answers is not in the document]
ConorTighe1995: If this is the case should we add `if (process) delete process._id` in here too? On one hand a function should only really do one thing and thats it but the `if` appears a lot in our code and is there a case where we would ever want to find a process and expose the db _id?

lib/utils.js
```diff
@@ -86,11 +86,15 @@ exports.getIotSupremeUserPolicy = () => ({
...
-exports.exclude = (obj, keys) => (
-  Object.keys(obj)
-    .filter(k => !keys.includes(k))
-    .reduce((res, key) => Object.assign(res, { [key]: obj[key] }), {})
-)
+exports.exclude = (obj, keys) => {
+  const newObj = { ...obj }
+
+  keys.forEach(key => {
+    delete newObj[key]
```
ConorTighe1995, line 93: This will throw a runtime error if key does not exist, maybe a try catch around delete would be good here and log a warning.

services/process/src/process/triggers/switchPhase.js
```diff
@@ -16,16 +15,13 @@ const Process = require('../process')
...
+  const { processId, startTime, phaseId } = event
+
+  await connectAllModels()
 
-  const db = await mongo.getDB()
-  const process = new Process(db)
+  await Process.addPhase(processId, { phaseId, startTime })
 
-  await process.switchPhase(requestData)
+  return 1
```
O-Susanoo, line 24: is this required tough ?

lib/models/Process/Process.js
O-Susanoo: configurator deviceName is not the same as this one here

services/process/src/process/delete.js
```diff
@@ -1,23 +1,71 @@
...
+  await connectAllModels()
 
-  validateDelete(requestData)
+  const user = await User.find(requestingUserId)
+  assert(() => canFind(user, `User ${requestingUserId}`))
 
-  const db = await mongo.getDB()
-  const process = new Process(db)
+  const processes = await Process.findMany(processesIds)
+  const validate = validateProcessCanBeDeletedByUser.bind(null, requestingUserId)
```
O-Susanoo, line 64: i guess if you have made an inline function here would be much more clearer something like 
```
validate = (process) =>  validateProcessCanBeDeletedByUser(requestingUserId, process)
```

services/process/src/process/grantAccess.js
```diff
@@ -1,25 +1,61 @@
...
+    () => userHasProcessActionAccess(device, process, requestingUserId, PROCESS_ACTIONS.SHARE),
+  ])
+
+  const userIdsWithProcessAccess = ProcessAccess.getUsers(process.access)
+  assert(() => isNotPresent(
+    userIdsWithProcessAccess,
+    user.id,
+    undefined,
+    'User is already present on this process',
+  ))
```
O-Susanoo, line 50: this node can be applied to all , since we are doing this rework , i think we could have made two things :
-ValidateSchema function which is the usual one
-ValidateRequest function which will always have the assert logic in it , because this pattern is repeated everywhere basically

lib/joiUtils.js
DanielSeehausen: it is a safety net as this is used for recording creation dates/timestamps/etc. 
if a negative number is provided it is a mistake and this should help early alert that

services/process/test/process/create.test.js
DanielSeehausen: ++ to that

lib/tasks/users/deleteAllUserRelatedData.js
DanielSeehausen: image
I can't follow the links

lib/models/Device/Processes/DeviceProcesses.js
DanielSeehausen: I wanted to make sure this scoured all places a references could be hanging out and stale
if the device ref was removed from the process, for example, this will still find it!

lib/models/Process/Process.js
DanielSeehausen: yes this was a rebase issue thanks. I will get rid of it

lib/models/Process/Process.js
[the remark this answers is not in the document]
DanielSeehausen: @ConorTighe1995 
I don't believe we ever expose the internal mongodb _id. I don't know if its an issue or not. There are definitely places where we do other queries and don't parse it out.

Re: your comment in adding it into here, in the diff I see it is in here already. did you mean we should remove it or add it somewhere else?

@vdiezel i'm not sure what you mean - is there something else this is a reference to that should be abstracted into this `find` method? is this about the `delete`? maybe you mean putting it in the mongo query `findOne` to remove the _id automatically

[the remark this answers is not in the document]
DanielSeehausen: or are you saying maybe we don't need a `find` at all?
I think a lookup interfaces is definitely required. While this is simple for now because its just by ID, its important to have a standard lookup interfaces so that you only need to change one place if some way we look stuff up changes.

.tmuxinator.yml
```diff
@@ -23,3 +23,5 @@ windows:
           - cd services/uIConfig
         - eventTrigger:
           - cd services/eventTrigger
+        - dataPoints:
+          - cd services/eventTrigger
```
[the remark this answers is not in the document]
DanielSeehausen: oh shoot good eye!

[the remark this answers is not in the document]
DanielSeehausen: fixed

lib/models/Process/Process.js
```diff
@@ -69,8 +161,54 @@ class Process {
...
+        as: 'devices',
+      } },
+    ]).next()
+
+    const device = process ? process.devices[0] : null
+
+    return { process, device }
+  }
+
+  async findProcessesWithDevices(processIds) {
```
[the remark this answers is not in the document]
DanielSeehausen: are you proposing this is `findProcessesWithDevice` instead of `...Devices`?
currently, this returns a single device affixed to every process, not an array.
If the name naturally says to you 'this is many processes, each with many devices', then im game for changing it. Ill do it now

lib/models/Process/Users/ProcessUsers.js
```diff
@@ -0,0 +1,26 @@
...
+  static async findUsers(process) {
+    /* Note: the user-process join is currently duplicated:
+     * user.process [...array of user ids]
+     * process.access (interface into users with user rights on a process)
+     *
+     * a user should always be in the other of they are in one
+     *
+     * We should consider addressing this at some point
+     * For now, this lookup defaults to the access array
+     */
```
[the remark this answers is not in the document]
DanielSeehausen: which totally makes sense and I share your fear
we could always make a `ProcessUsers` collection in the future

lib/utils.js
DanielSeehausen: image

@ConorTighe1995 perhaps this is not what you meant, but I tried to recreate it and did not see it throwing here. 

Just in case this was a version thing, here it is on 12 now as well, which hambda was updated to:
image

services/process/src/process/triggers/switchPhase.js
DanielSeehausen: good eye. no it's definitely not as I understand, as no consumers do anything with the responses for the time being.

services/process/src/process/delete.js
DanielSeehausen: yes you are 100% on that one I don't know what I was thinking.
I just updated

services/process/src/process/grantAccess.js
DanielSeehausen: I think further abstracting things like validating request schemas definitely makes sense at some point. Outside of some convenient middleware, I am unsure what the path forward in abstracting that would be. What are you proposing?

Re: the assertions, I am unsure how to abstract those further. How would it look with what you are proposing? What I find hard to abstract are the following two things: 
1. Different routes/lambda handlers have different specific assertions required, which means if they line up with another routes assertions it's only by chance, and assertions on a request (e.g. permissions) are subject to change
2. in terms of reading what the method does, its pretty helpful for a dev to read down a list of assertions. Names would get pretty long if we combined them!

Maybe these concerns aren't noteworthy if we have the right implementation. If you have something specific in mind let me know - I am all ears

lib/utils.js
ConorTighe1995: Your right actually, I tested it in a online compiler and that gave me an error  but seeing that it works in the node cmd Im wrong, never mind!

lib/models/Process/Process.js
```diff
@@ -1,62 +1,153 @@
...
+    return Process.processesDB.find(
+      { id: { $in: processIds } },
+    ).sort({ createdAt: -1 }).toArray()
   }
 
-  static addPhase(process, phase) {
-    process.progression.push(phase)
+  static async delete(processId) {
+    // NOTE: if you are here, you might want to remove the process and all related data. see:
+    // lib/tasks/processes/deleteAllProcessRelatedData.js
```
ConorTighe1995, line 132: I like the little note here, definitely will help in the future

services/process/src/process/grantAccess.js
ConorTighe1995: Just to be clear you guys are talking about  `userHasProcessActionAccess` here? or You want to run lines 37-50 on every request?

services/process/src/process/modifyAccess.js
```diff
@@ -1,24 +1,62 @@
...
+} = require('validation-lib')
 const { success } = require('http-responses')
 const HTTPDefaultMiddleware = require('middleware/HTTPDefaultMiddleware')
-const Process = require('./process')
 
+const validateModifyAccess = require('../joi/process').modifyAccess
+
+
+// NOTE/TODO: do we still need the old group at all?
+// Users can only be present in a single group at a time
```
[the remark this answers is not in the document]
DanielSeehausen: thank you! - I will remove this in the next pr that's already branched off this

services/process/src/process/update.js
```diff
@@ -1,25 +1,50 @@
...
+    pathParameters: { id: processId },
+    body: updateData,
+  } = event
+
+  validateUpdate({ requestingUserId, processId, updateData })
+
+  await connectAllModels()
+
+  const { process, device } = await Process.findWithDevice(processId)
+  //console.log(process, device)
```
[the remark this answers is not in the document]
DanielSeehausen: thank you! - I will remove this in the next pr that's already branched off this

PR #452, opened 2020-05-13
URGENT: fix for user creation

vdiezel:
We had a small bug that led to users being created without certain fields on their user model in the database. It broke the app and my heart.

Review just the last commit here (I branched off from daniel's branch for easier merging)


vdiezel:
replaced by : https://github.com/ospin-web-dev/hambda/pull/453

PR #455, opened 2020-05-15
448 allow custom conversions on stream config

### LINKED ISSUE #448: Feature: RPM to flow/time unit conversion - allow customConversions in the stream config
To support unit conversion from rpm to ml/min we need a way to store the user input for the ratio between rotation and ml. This should go into a new field on the stream configuration named `customConversions` and should be something like this: 

```js
customConversions: [ { to: 'ml/min', ratio: 0.4 }]
```

adjust the Joi schema for updateDevice to allow this new property on the stream configuration and add tests for it.

vdiezel:
closes #448


services/device/test/device/update.test.js
```diff
@@ -82,6 +82,7 @@ describe('Device Class', () => {
                 slope: 1.2,
               },
               eventTriggers: [],
+              customConversions: [{ ratio: 0.4, to: 'ml/min' }],
```
ConorTighe1995, line 85: This may be a stupid question but why is it 0.4 here?

vdiezel: just a random number a user might have put in - has no meaning at all

vdiezel:
> this is looking good
> 
> Might also be time to make a device stream SCHEMA. Only this could also work for now, but I think we may want our data to be verified outside of routes themselves

I agree. I will postpone this to when we implement the device model!

vdiezel:
replaced by #460

PR #461, opened 2020-05-18
425 reduce process payload

### LINKED ISSUE #425: Fetching Device Processes Speed Up
### Description
Ever increasing amount of processes is slowing down the user experience in the web application when we fetch device processes. (almost 10 seconds on celltainer when I tried just now! Some of that may be from a lambda cold start)

related to: ospin-web-dev/REACTor#651

---

### Proposed Solution(s)
There are several things we can do to improve this experience (fetching more processes on scroll/pagination), but I am only proposing to do one right now:

Analyze where in our backend/the network communication the slow down is in GET device processes (is it in the mongo call, is it the size of the response getting to the client, is it something else?)

Optimize the device processes look up depending on where the slow down is caused.

One direction this might need to go (if it's a response size issue) is to send light process data when we fetch device processes. This will require investigation on the front end to ensure everything still works as intended, and that whenever a /device/id/process/id route is hit, that the whole process is loaded.

If front end work is required, update this ticket with the required work!:
ospin-web-dev/REACTor#651

---

### Deliverable(s)
As a user:
Given how many processes devices on dev have right now,
fetching device processes is a faster operation than it is now

This does not include a pagination implementation!

---

image

===== Update ===== 
the end point should return this fields 

```
{
  id,
  name,
  comment,
  createdAt,
  createdBy,
  cloneHistory,
  access,
  pinned,
  ownerId,
  startedAt,
  startedBy,
  finishedAt,
}
```

DanielSeehausen:
This is ready to go - it updates both list processes as well as list device processes

There is a lot of fluff and it might be easiest to go through files changes and just collapse all the ones that have to do with the test refactor on connect all models

will be closed by #471 

### What this does

other than making a lot of noise in the tests, it leaves in some helper methods that I used while updating data, removes some deprecated script stuff, and (most importantly) changes the processes list (and device process list) response to be a lot slimmer.


DanielSeehausen:
image

:O

DanielSeehausen:
Here are some speeds/sizes on staging for the celltainer device loading processes directly from
`https://staging.ospin-app.com/device/9ef4f0b3-977b-4d36-9760-bba0b789bb0c/processes` 
And from Athens on localhost with the process list code deployed loading from:
`http://localhost:3000/device/a8207dab-43ab-44be-93eb-308a60df1f29/processes`

### Payload sizes
(with some back of the envelope math ignoring http response header size)
Celltainer was: 386 processes at 3.9 MB total (.98 MB per 100 processes)
Athens was:      110 processes at  1.0 MB total (1.1 MB per 100 processes)

These numbers seem very surprising from 1000ft away. I will need to double check all of this tomorrow and make sure everything is deployed correctly. Hopefully the numbers aren't correct or I am doing something wrong here/not understanding, because we wanted to see a much bigger reduction in payload size. 

#### Times

Celltainer (old code):
4:25
2:86
3:13
2:63
2:93
4:93
7.82
1.66
1.56!
1.60!
1.52!
1.41!
1.66!
1.44!

image

Athens (new code with ):
.629 (x3 2.1)
.473 (x3 1.3)
.471 (x3 1.3)
.417 (x3 1.2)

It really needs to be tested with the same code payload size to verify, as the first test is 3.5 times larger of a payload

package.json
```diff
@@ -13,8 +13,9 @@
     "testAll": "nps testAll",
     "generateDocs": "nps generateDocs",
     "testAllWithCoverage": "nps testAllWithCoverage",
-    "runScriptDev": "node -r dotenv/config scripts/runner dotenv_config_path=env.dev.yml",
-    "runScriptProd": "node -r dotenv/config scripts/runner dotenv_config_path=env.prod.yml"
+    "seedLibFiles": "cp -rf ./lib/* ./node_modules/",
```
DanielSeehausen, line 16: this is used when running scripts (to keep the import/exports consistent with what a dev experiences in the services themselves)

lib/models/Device/Processes/DeviceProcesses.js
```diff
@@ -1,5 +1,9 @@
 class DeviceProcesses {
```
vdiezel, line 1: wouldn't this class qualify to extend the `Process` model? Then `findProcesses(deviceId)` could be replaced by `DeviceProcesses.findByQuery({ deviceId })`

lib/models/Process/Process.js
```diff
@@ -114,13 +115,29 @@ class Process {
...
+
   static async find(processId) {
     const process = await Process.processesDB.findOne({ id: processId })
     if (process) delete process._id
 
     return process
   }
 
+  static async findByQuery(query) {
+    return this.processesDB.find(query).toArray()
```
vdiezel, line 138: `this` -> `Process`

lib/models/connectAllModels.js
```diff
@@ -10,12 +11,15 @@ const CONNECTABLE_MODELS = [
...
+  EventRevoker,
   Process,
   UIConfig,
   DataPoints,
 ]
 
 if (process.env.NODE_ENV === 'test') {
+  // If this is moved out of this confitional, we will need to ensure all services that use this file alse have dependencies it goes on to require installed
+  // There is certainly a better way to do this!
   const Seeder = require('../test-helpers/seeder/Seeder')
```
vdiezel, line 23: The cleaner way would be simly not to connect Seeder at all here, but explicitly in the test files instead.

services/device/src/serializers/serializeListDeviceProcesses.js
```diff
@@ -0,0 +1,21 @@
...
+  'startedAt',
+  'finishedAt',
+  'pinned',
+  'cloneHistory',
+  'comment',
+]
+
+module.exports = processes => (
+  processes.map(process => pick(...INCLUDED_KEYS)(process))
+)
```
vdiezel, line 21: I am wondering if we can make the performance even better by filtering the properties we want already in the mongo query.

services/process/test/process/list.test.js
```diff
@@ -70,6 +71,37 @@ describe('listProcesses', () => {
...
+
+    const processCount = 5
+    await createProcesses(user, device, processCount)
+
+    const { res, body } = await fireListProcessesEvent(user.id)
+    expect(res.statusCode).toBe(200)
+
+    const keys = body.reduce((acc, process) => (
+      new Set([ ...acc, ...Object.keys(process) ])
+    ), new Set())
```
vdiezel, line 85: If I read this the right way, you are creating a set of all possible keys from all processes. Shouldn't every process have the same keys anyway?

services/device/src/serializers/serializeListDeviceProcesses.js
ConorTighe1995: Could be worth a try

lib/models/Device/Processes/DeviceProcesses.js
DanielSeehausen: good idea - I remove this new method from `DeviceProcesses` and will use the `Process.findByQuery` instead

lib/models/Process/Process.js
DanielSeehausen: updated

lib/models/connectAllModels.js
DanielSeehausen: i definitely agree with that. That sounds like a good direction to head going forward. 
Because of how closely tied this is with a lot of tests still, I'd propose we leave it for this PR and update it next chance we have

services/device/src/serializers/serializeListDeviceProcesses.js
DanielSeehausen: I suspect it would increase performance! I wrestled with this for a different reason - I was wondering whether this should be a general processes interface method or something that is specific to serialization for a response.

It probably makes sense to standardize this on the interface (which means we could do it within Mongo). With this i'd also propose we leave it for now and update next time it is touched (which it will be, as people have ever increasing processes).

lib/models/connectAllModels.js
vdiezel: Ok! I've read this after I requested the changes, so simply ignore there.

services/process/test/process/list.test.js
DanielSeehausen: This is to account for processes that do not have keys (maybe a key was not successfully set to `null`, for example, and is just lacking from the process object). The test wouldn't fail in this case. Upon considering this further, I don't know why I left it able to pass when a process didn't have an expected key. I will update.

ConorTighe1995: But wouldnt looping through every single process still be slower then throwing the keys in a set to remove the duplicates and checking that? I assume hes going for performance here.

services/process/test/process/list.test.js
```diff
@@ -70,6 +71,37 @@ describe('listProcesses', () => {
...
+      'createdAt',
+      'name',
+      'access',
+      'startedBy',
+      'startedAt',
+      'finishedAt',
+      'pinned',
+      'cloneHistory',
+      'comment',
+    ].sort())
```
ConorTighe1995: Do we need to perform 2 sorts here? Would we get away with the following to further improve performance:
https://jestjs.io/docs/en/expect#expectarraycontainingarray

DanielSeehausen:
taken up in a merged PR #471

PR #463, opened 2020-05-19
adds test to removal of UI configs

### LINKED ISSUE #443: `lib/tasks/users/deleteAllUserRelatedData.js` should also delete user uIConfigs!
### Description

`lib/tasks/users/deleteAllUserRelatedData.js` should also delete user uIConfigs!
I totally forgot to add that!

---

### Proposed Solution(s)
update the task! update any uIConfig helper methods required
provide tests!

Is there anything else we missed that is user specific?

vdiezel:
It was actually implemented already. I just added test. I have a question regarding one line of code though.


lib/models/UIConfig/UIConfig.js
```diff
@@ -72,6 +72,7 @@ class UIConfig {
       () => canFind(device, `Device ${deviceId}`),
     ])
 
+    // is this necessary?
     UIConfig.schema(userId, deviceId, configs)
```
vdiezel, line 76: line 76 is there to validate the payload for an update or create of a UIconfig. If you check the code further down, it calls either `UIConfig.create`  or `UIConfig.update`. Both of them also assert the schema, either via the `schema` call in create or the `Joi.attempt()` in update. Was line 76 added to cover the case of an update, because `Joi.attempt()` works differently?

I was just wondering because if I pass in an empty object as the `configs` parameter to `UIConfig.updateOrCreate` it will throw a validation error in line 76 and complaining about wrong schema for the `configs`. This is because the default configs are set in `UI.create`. 

As a developer I might think I can pass an empty object for configs to `updateOrCreate` and then it sets the default parameters, like it does in `UIConfigs.findOrCreate()`.

So my question: Is line 76 necessary at all?

DanielSeehausen: if both update and create assert the schema then this can definitely go!

services/user/test/user/delete.test.js
```diff
@@ -99,6 +100,28 @@ describe('DELETE user', () => {
...
+
+      const configs = await UIConfig.findAll()
+      expect(configs.length).toBe(2)
+
+      await fireDeleteUserEvent(user.id, user.id)
+
+      const configsPostDelete = await UIConfig.findAll()
+      expect(configsPostDelete.length).toBe(0)
+    })
+
```
DanielSeehausen, line 124: NOICE

vdiezel:
replaced by https://github.com/ospin-web-dev/hambda/pull/471

PR #468, opened 2020-05-19
433 demo user route for creating demo accounts

### LINKED ISSUE #433: H2: Create new route user/demo for demo accounts
### Description
When a user attempts to use a demo versions of the application then a route called user/demo should be hit which will handle the creation of demo accounts and logging in of existing demo accounts.

---

### Proposed Solution(s)
Add a new route lambda route/handler in the user service that will create + authorize a demo user in cognito. 

Below is a link with 2 different solutions on how we could achieve guest login: [link here](https://github.com/aws-amplify/amplify-js/issues/432#issuecomment-372889576). 

Personally I think the hard coded password option might be the best for an MVP. We can track the guest accounts session by saving the userId to local storage in the browser or as a browser cookie. This will cut down on account creation and allow guest users to resume the demo session if they close the browser or refresh. We will use the `user.role === 'demo'?` flag to limit features on the front end like change password in later work. 

Investigate unauth users in congnito!

---

### Deliverable(s)

- [ ] New user/demo route
- [ ] Creates user if one does not exist (try extending the `User` interface into a `DemoUser` interface!)
- [ ] User will have user.role: 'demo'
- [ ] Returns a User object with an access_token (the user object will self describe as a demo user with the using the user type property. Re: the demoId, what will that provide us? Are we able to use the userId?)

ConorTighe1995:
This route was created with the intention of generating a new demo account be implementing a pre-sign up step before the create account route used in a regular sign up. This route assigns the user attributes and a flag for cognito to bypass the email verification.

The route will trigger a callback to cognito which will then proceed with the signup process as normal, creating a new cognito account and user model in Mongodb.  Username on these accounts have timestamps attached... Maybe we could later use this for cleaning up demo accounts?

There is a hard coded guest password that is replaceable in prod with a secret.

closes #433


services/user/serverless.yml
```diff
@@ -325,6 +325,17 @@ functions:
       DB_USER: ${{env:DB_USER}}
       DB_PASSWORD: ${{env:DB_PASSWORD}}
       DB_NAME: ${{env:DB_NAME}}
+  createDemoUser:
+    handler: src/user/createDemoUser.main
+    role: AWSLambdaLoggingRole
+    events:
+      - cognitoUserPool:
+          pool: OspinUserPool
+          trigger: PreSignUp
```
ConorTighe1995, line 334: This tells cognito to run createDemoUser before the regular create.

services/user/src/user/createDemoUser.js
```diff
@@ -0,0 +1,24 @@
+const connectAllModels = require('models/connectAllModels')
+const DemoUser = require('models/DemoUser/DemoUser')
+const HTTPDefaultMiddleware = require('middleware/HTTPDefaultMiddleware')
+const { success } = require('http-responses')
+
+async function createDemoUser(event, context, callback) {
```
ConorTighe1995, line 6: This file is the star of the show, the meat!

services/user/src/user/createDemoUser.js
```diff
@@ -0,0 +1,24 @@
...
+  const newUserData = {
+    userName: `guest-${new Date().getTime()}`,
+    role: 'demo', // currently restricted to just 'user'
+    password: process.env.GUEST_PASSWORD || 'GuestPassword12345',
+  }
+
+  // Set the user pool autoConfirmUser flag after validating the email domain
+  event.response.autoConfirmUser = true
```
ConorTighe1995, line 14: This is the flag that will let us bypass the email verification.

lib/test-helpers/test-helper.js
```diff
@@ -2,16 +2,19 @@ exports.assertContainsAll = (string, array) => {
   array.forEach(element => expect(string).toContain(element))
 }
 
-const promisify = async (handler, event, context) => (
+const promisify = async (handler, event, context, callback) => (
```
ConorTighe1995, line 5: We had to modify these to make the callback for createDemoUser work.

DanielSeehausen:
is this still a WIP? won't review if so!

lib/test-helpers/test-helper.js
O-Susanoo: Not anymore

services/user/src/user/createDemoUser.js
O-Susanoo: We dropped this approch

services/user/src/user/createDemoUser.js
O-Susanoo: Not anymore this file is more now like a sidekick

services/user/serverless.yml
O-Susanoo: We dropped this approach we are now manually creating the cognito user

O-Susanoo:
We are only missing tests here and everything should be fine after that

ConorTighe1995:
> is this still a WIP? won't review if so!

Its still WIP, but we shouldn't be much longer as its mostly done.

services/user/serverless.yml
```diff
@@ -325,6 +354,20 @@ functions:
       DB_USER: ${{env:DB_USER}}
       DB_PASSWORD: ${{env:DB_PASSWORD}}
       DB_NAME: ${{env:DB_NAME}}
+  createDemoUser:
```
ConorTighe1995, line 357: This is the lambda function config to trigger demo user creation

ConorTighe1995:
This is now ready for review.

lib/models/DemoUser/DemoUser.js
```diff
@@ -0,0 +1,36 @@
+const uuidv4 = require('uuid').v4
+const mixin = require('functional-helpers/mixin')
+const Connectable = require('models/Connectable')
+const CognitoAPI = require('cognito/CognitoAPI')
+
+class DemoUser {
+
+  static async create(timestamp) {
+    const password = 'GuestPassword12345'
```
vdiezel, line 9: this should be an environmental variable

lib/models/DemoUser/DemoUser.js
```diff
@@ -0,0 +1,36 @@
...
+    const userName = `guest-${timestamp}`
+    const user = {
+      id: uuidv4(),
+      role: 'demo',
+      userName,
+      email: `${userName}@ospin.de`,
```
vdiezel, line 15: why do we need this?

vdiezel:
> The route will trigger a callback to cognito which will then proceed with the signup process as normal, creating a new cognito account and user model in Mongodb. Username on these accounts have timestamps attached... Maybe we could later use this for cleaning up demo accounts?

A more secure and reliable way would be to use the role. Via the asscoiation of the Cognito user sub and the user id in the database we can delete them in both places.

services/user/src/user/createDemoUser.js
```diff
@@ -0,0 +1,13 @@
+const connectAllModels = require('models/connectAllModels')
+const DemoUser = require('models/DemoUser/DemoUser')
+const HTTPDefaultMiddleware = require('middleware/HTTPDefaultMiddleware')
+const { success } = require('http-responses')
+
+async function createDemoUser(event, context) {
+  await connectAllModels()
+  const timestamp = new Date().getTime()
```
vdiezel, line 8: What are you planning to use the timestamp for? If its for uniqueness, then I would use some kind of uuid (doesnt have to be uuidv4). 
If you want to use the time itself: Every user has a `createdAt` property in the database. Also the MongoDB ObjectID (`_id`) has the creation time actually encoded in it https://docs.mongodb.com/manual/reference/method/ObjectId.getTimestamp/

DanielSeehausen:
I have comments in the delete stale users pr that pertain to this. i will make a PR for those requested changes to this branch!

services/user/src/user/createDemoUser.js
DanielSeehausen: this has been updated here: https://github.com/ospin-web-dev/hambda/pull/477

lib/models/DemoUser/DemoUser.js
DanielSeehausen: removed here: https://github.com/ospin-web-dev/hambda/pull/477

more specifically: https://github.com/ospin-web-dev/hambda/pull/477/files#diff-82bf59c1d9dfd2d4a85f806a2d87c34fR19

ConorTighe1995:
@vdiezel We could hash the guest password, we would need to add a prelambda function that reverses the hashing of the password for login, ill look into it now.

lib/models/User/DemoUser.js
```diff
@@ -0,0 +1,29 @@
+const User = require('models/User/User')
+const uuidv4 = require('uuid').v4
+
+class DemoUser extends User {
+
+  // TODO: move this to env variable and update env files on github org wiki and alert team in slack to download new env files (and remove from here)
```
DanielSeehausen, line 6: now that this is referencing an env var, this comment can go!

services/user/src/user/createDemoUser.js
```diff
@@ -0,0 +1,31 @@
...
+
+  // confirmUser will call the post confirmation trigger and create a user in the Database
+  // and it looks like this runs synchroniously, so after line 18 the user will be in the database!
+  await CognitoAPI.confirmUser(userName)
+
+  const user = await DemoUser.find(UserSub)
+
+  // do your demo user specific logic here, like set the role to demo and add devices and processes
+
+  // TODO: reference env var instead of DemoUser.GUEST_PASSWORD once its moved to env var
```
DanielSeehausen, line 24: this comment can go now!

services/user/test/user/createDemoUser.test.js
```diff
@@ -0,0 +1,57 @@
...
+      expect(CognitoAPI.createUser).toHaveBeenCalledTimes(1)
+    })
+
+    it('should return 200 and a user object and a password that is the same as the demo user password env var', async () => {
+      const { res: { statusCode }, body: returnedUser } = await fireCreateDemoUserEvent()
+      const dbDemoUser = await DemoUser.find(returnedUser.id)
+
+      expect(statusCode).toEqual(200)
+      expect(returnedUser).toEqual(expect.objectContaining(dbDemoUser))
+      // TODO: move this to check a test env var instead of the class getter (we will need to start loading a test env var file!)
```
DanielSeehausen, line 53: this comment can be removed

services/user/src/user/createDemoUser.js
```diff
@@ -0,0 +1,31 @@
...
+    userName, DemoUser.GUEST_PASSWORD, userName + DemoUser.GUEST_EMAIL_DOMAIN,
+  )
+
+  // confirmUser will call the post confirmation trigger and create a user in the Database
+  // and it looks like this runs synchroniously, so after line 18 the user will be in the database!
+  await CognitoAPI.confirmUser(userName)
+
+  const user = await DemoUser.find(UserSub)
+
+  // do your demo user specific logic here, like set the role to demo and add devices and processes
```
DanielSeehausen, line 25: I believe this still may need to be done

services/user/src/user/create.js
```diff
@@ -18,13 +19,18 @@ async function create(event, context) {
...
   await connectAllModels()
 
+  if (newUserData.role === 'demo') {
+    await DemoUser.create(newUserData)
+    context.succeed(event)
+  }
+
   await User.create(newUserData)
 
   context.succeed(event) // triggers lambda to close via cognito request method
```
DanielSeehausen, line 36: could be slightly simpler expressed to avoid the multiple returns by:
```js
const UserType = newUserData.role === 'demo' ? DemoUser : User
await UserType.create(newUserData)

context.succeed(event)
```

definitely not blocking, just wanted to provide a way to express is shorthand while avoiding the eslint warning it gives you for invoking stuff in a ternary

services/user/src/user/createDemoUser.js
```diff
@@ -8,23 +8,20 @@ async function createDemoUser() {
...
   // confirmUser will call the post confirmation trigger and create a user in the Database
   // and it looks like this runs synchroniously, so after line 18 the user will be in the database!
   await CognitoAPI.confirmUser(userName)
 
-  const user = await DemoUser.find(UserSub)
-
-  // do your demo user specific logic here, like set the role to demo and add devices and processes
-
-  return success({ ...user, password: DemoUser.GUEST_PASSWORD })
+  return success({ userName, password: DemoUser.GUEST_PASSWORD })
```
DanielSeehausen: it looks like we removed the lookup:
`const user = await DemoUser.find(UserSub)`

that seemed like a pretty good sanity check to have in place (to ensure line 22 above passes). any way we can keep that or did we have to get rid of it for a rason?

services/user/src/user/createDemoUser.js
```diff
@@ -8,23 +8,20 @@ async function createDemoUser() {
...
 
-  const { UserSub } = await CognitoAPI.signUpUser(
+  await CognitoAPI.signUpUser(
     userName,
     DemoUser.GUEST_PASSWORD,
     email,
+    'demo',
   )
 
   // confirmUser will call the post confirmation trigger and create a user in the Database
```
DanielSeehausen, line 20: comment is no longer accurate re: line number and maybe misleading. line number itself can be removed because you reference the method `confirmuser`

services/user/test/user/createDemoUser.test.js
```diff
@@ -38,13 +38,16 @@ describe('DELETE user', () => {
...
       await fireCreateDemoUserEvent()
 
-      expect(CognitoAPI.createUser).toHaveBeenCalledTimes(1)
+      expect(CognitoAPI.signUpUser).toHaveBeenCalledTimes(1)
+      expect(CognitoAPI.confirmUser).toHaveBeenCalledTimes(1)
     })
 
-    it('should return 200 and a user object and a password that is the same as the demo user password env var', async () => {
+    // TODO fix status code error once creation of demo user in done
+    xit('should return 200 and a user object and a password that is the same as the demo user password env var', async () => {
```
DanielSeehausen, line 50:
```suggestion
    it('should return 200 and a user object and a password that is the same as the demo user password env var', async () => {
```

for once used the suggest a change feature

DanielSeehausen: i see the TODO above. does this PR not solve this?

DanielSeehausen:
updated on master through UI here

services/user/test/user/createDemoUser.test.js
ConorTighe1995: We added the TODO, the test broke because of the implementation changes

services/user/src/user/createDemoUser.js
ConorTighe1995: Its done in DemoUser, ill remove the comment now, we still need to add the virtual processes though

services/user/src/user/createDemoUser.js
ConorTighe1995: We removed that as we only needed to return the userName which we already had and the call to the db was unnecessary

lib/models/User/DemoUser.js
```diff
@@ -9,16 +11,37 @@ class DemoUser extends User {
 
   static get GUEST_USERNAME() { return `demo-user-${uuidv4().substr(0, 8)}` }
 
+  static get SCHEMA() {
+    return super.SCHEMA.keys({
```
DanielSeehausen, line 15: extends super shema pretty cool JOI

PR #469, opened 2020-05-20
306 add logs when deleting processes

### LINKED ISSUE #306: Deleting processes should create device log
Deleting processes goes by silently currently. Instead we should create a device log that contains the information about deleted processes: name, id and user who deleted it.

vdiezel:
- Deleting processes will now create log entries in the device log

\private
Bildschirmfoto von 2020-05-20 13-15-31

Every process gets its own log entry, this way we could query them by processId


services/process/src/logger/Logger.js
```diff
@@ -0,0 +1,29 @@
...
+    return `User ${user.userName} deleted process ${process.name}`
+  }
+
+  static deleteProcess(user, process) {
+    return {
+      deviceId: process.deviceId,
+      processId: process.id,
+      level: 'information',
+      subject: 'device/operation',
+      message: Logger.createDeleteProcessText(user, process),
```
ConorTighe1995, line 17: Nothing necessarily wrong here but will `createDeleteProcessText` appear somewhere else in the future? If not this might be leaning towards boilerplate code, you could just pass `User ${user.userName} deleted process ${process.name}` straight to message.

vdiezel: you are right. I changed it

vdiezel:
merged togfether with this: 
https://github.com/ospin-web-dev/hambda/pull/471

PR #471, opened 2020-05-20
425, 443, 306, 434 part 1

### LINKED ISSUE #306: Deleting processes should create device log

### LINKED ISSUE #425: Fetching Device Processes Speed Up

### LINKED ISSUE #443: `lib/tasks/users/deleteAllUserRelatedData.js` should also delete user uIConfigs!

DanielSeehausen:
part 1 seeks to make a create device functionality independent of authorization
this will be re-used in the cloning of virtual devices
this does not close 434 - the next ticket will

closes #425 
closes #443 
closes #306


lib/models/Device/DemoDevice.js
```diff
@@ -0,0 +1,11 @@
+const Device = require('models/Device/Device')
+
+
+class DemoDevice extends Device {
+
+  //static async create(deviceId) {
+  //}
+
+}
```
DanielSeehausen, line 9: this will be used in the next PR

lib/models/Device/Device.js
```diff
@@ -1,22 +1,107 @@
...
+const DeviceConfig = require('models/Device/DeviceConfig/DeviceConfig')
+const User = require('models/User/User')
 const DeviceAccess = require('ResourceAccess/DeviceAccess')
 const Connectable = require('models/Connectable')
 const mixin = require('functional-helpers/mixin')
 
 
 class Device {
 
+  // TODO: make extended class for bioreactor
```
DanielSeehausen, line 15: note for next PR

lib/models/Device/DeviceConfig/DeviceConfig.js
```diff
@@ -0,0 +1,33 @@
...
+  static get SCHEMA() {
+    return Joi.object({
+      global: Joi.object().default({}),
+      streams: Joi.array().items(
+        Joi.object({
+          streamId: Joi.string().required(),
+          defaultUnit: Joi.string().required(),
+        }),
+      ).default([]),
+    })
```
DanielSeehausen, line 15: if anyone is able to confirm this streams structure I would be very grateful

lib/models/Device/Device.js
```diff
@@ -1,22 +1,107 @@
...
+      createdAt: unixTimeDefaultNull(),
+      createdBy: Joi.string().pattern(UUIDV4_REGEX).default(null).allow(null),
+      name: Joi.string().min(1).max(45).required(),
+      isVirtual: Joi.boolean().required(),
+      type: Joi.boolean().valid(...Device.DEVICE_TYPES).required(),
+      streams: Joi.array().default([]),
+      ownerId: Joi.string().pattern(UUIDV4_REGEX).default(null).allow(null),
+      hardwareVersion: Joi.number().integer().default(0),
+      processes: Joi.array().default([]),
+      groupNameAlias: Joi.array().default([]),
```
vdiezel, line 36: `groupNameAlias` is unused. I created a new ticket: 
https://github.com/ospin-web-dev/hambda/issues/472

lib/models/Device/Device.js
```diff
@@ -1,22 +1,107 @@
...
   static async find(deviceId) {
     const device = await Device.devicesDB.findOne({ id: deviceId })
     if (device) delete device._id
 
     return device
   }
 
+  static async findAll() {
+    const devices = await Device.devicesDB.find().toArray()
+    if (devices) devices.forEach(device => delete device._id)
```
vdiezel, line 81: if you want to be 100% here, you would need to check for `devices.length` because an empty array is still truthy.  Your code doesn't break - but the condition will be always true here

lib/models/Device/Device.js
```diff
@@ -1,22 +1,107 @@
...
+
+    return devices || []
```
vdiezel, line 83: same here, no need for the default fallback as devices will always be always at least an empty array.

lib/models/Device/DeviceConfig/DeviceConfig.js
vdiezel: here you go: https://github.com/ospin-web-dev/hambda/blob/master/services/device/src/joi/device.js#L129

So you are missing a couple of keys ;)

The keys 'nameAlias', 'customDefault' and 'calibration' are not used yet but are there in preperation of moving the top level properties 'streamNamaAliases', 'streamDefaults' and 'calibration' into the config of the device.
The important ones are customConversions and eventTriggers.

lib/models/Device/DeviceConfig/DeviceConfig.js
```diff
@@ -0,0 +1,33 @@
...
+
+  static schema(configData) {
+    return Joi.attempt(configData, DeviceConfig.SCHEMA)
+  }
+
+  static create() {
+    const newConfig = {
+      global: {},
+      streams: [],
+    }
```
vdiezel, line 26: we had a nice pattern in the UI config of having a `DEFAULT_CONFIG` that will work here, too!

services/device/src/device/create.js
```diff
@@ -1,26 +1,46 @@
...
+
+  const user = await User.find(requestingUserId)
+  assertMany([
+    () => isMe(requestingUserId, newDeviceData.ownerId), // for now, users can only create devices with themselves as owners
+    () => hasDeveloperAccess(user),
+  ])
 
-  validateCreate(requestData)
+  const device = await Device.create(newDeviceData, requestingUserId)
+  // Device.create automatically adds the _owner_ as an admin -- this endpoint ensures the requestor (in case they are different from the owner), is also added as an admin
```
vdiezel, line 37: :+1:

PR #473, opened 2020-05-25
437 delete stale demo account

### LINKED ISSUE #437: H6: Stale demo accounts are deleted.
### Description
Demo accounts will build up over time with cognito and mongo getting filled with unused data. We need a way to clear these out manually or ideally automatically.

---

### Proposed Solution(s)
 A script might be needed here that runs based on a timestamp or heuristic value.

---

### Deliverable(s)
- [ ] Stop stale user accounts from building up

O-Susanoo:
---
daniel:

ready to go
closes #437 

The serverless file stuff I am unsure about - and didn't change anything there from the original PR


lib/models/DemoUser/DemoUser.js
```diff
@@ -0,0 +1,45 @@
+const uuidv4 = require('uuid').v4
+const mixin = require('functional-helpers/mixin')
+const Connectable = require('models/Connectable')
+const deleteAllUserRelatedData = require('tasks/users/deleteAllUserRelatedData')
+const CognitoAPI = require('cognito/CognitoAPI')
+
+class DemoUser {
```
DanielSeehausen, line 7: this file looks like a great start

I believe this should extend the `User` class and make use of `super` in at least `create`

lib/models/DemoUser/DemoUser.js
```diff
@@ -0,0 +1,45 @@
...
+    user.createdAt = timestamp
+    const { insertedCount } = await DemoUser.usersDB.insertOne(user)
+    const success = insertedCount === 1
+    if (success) delete user._id
+    return user
+  }
+
+  static async findUnActiveDemoUsers(timestamp) {
+    return DemoUser.usersDB.find({ createdAt: { $lt: timestamp } }).toArray()
+  }
```
DanielSeehausen, line 30: once extending the `User` class, you could use `findBy` to reduce direct database access

lib/models/DemoUser/DemoUser.js
```diff
@@ -0,0 +1,45 @@
...
+
+  static async removeUnActiveUsers(timestamp) {
+    const unActiveDemoUsers = await DemoUser.findUnActiveDemoUsers(timestamp)
+    return Promise.all(unActiveDemoUsers.map(async demoUser => {
+      await deleteAllUserRelatedData(demoUser.id)
+      return demoUser.id
+    }))
+  }
```
DanielSeehausen, line 38: this is extending a little beyond the scope of the demo user interface (base `User` already has `delete` on it and that should probably be enough once this is extending it). Whatever the service/task is that is doing the clean up, that is what should probably be responsible for doing the `deleteAllUserRelatedData`. The tasks can maybe be best thought of as higher level actions that delegate responsibilities down to the interfaces, instead of the other way around

services/user/src/user/createDemoUser.js
```diff
@@ -0,0 +1,13 @@
+const connectAllModels = require('models/connectAllModels')
+const DemoUser = require('models/DemoUser/DemoUser')
+const HTTPDefaultMiddleware = require('middleware/HTTPDefaultMiddleware')
+const { success } = require('http-responses')
+
+async function createDemoUser(event, context) {
+  await connectAllModels()
+  const timestamp = new Date().getTime()
+  const demoUser = DemoUser.create(timestamp)
```
DanielSeehausen, line 9: extending the `User` class should free you up from needing to do this iirc
I belie `DemoUser` needs to be awaited

test/lib/models/DemoUser/DemoUser.test.js
```diff
@@ -0,0 +1,59 @@
...
+      await DemoUser.create(timestamp)
+      expect(CognitoAPI.createUser).toHaveBeenCalledTimes(1)
+    })
+
+    it('should assign the correct username to the demo account', async () => {
+      const timestamp = 12345
+      // const demoUser = await DemoUser.create(timestamp)
+
+      // expect(demoUser.username).toBe('guest-12345')
+    })
```
DanielSeehausen, line 36: this test may not be finished

lib/models/DemoUser/DemoUser.js
```diff
@@ -0,0 +1,45 @@
...
+    const userName = `guest-${timestamp}`
+    const user = {
+      id: uuidv4(),
+      role: 'demo',
+      userName,
+      email: `${userName}@ospin.de`,
+      processes: [],
+      devices: [],
+    }
+    await CognitoAPI.createUser(user.userName, password)
```
DanielSeehausen, line 20: I believe this might be better suited in the lambda function instead of the model interface

lib/models/DemoUser/DemoUser.js
```diff
@@ -0,0 +1,45 @@
+const uuidv4 = require('uuid').v4
+const mixin = require('functional-helpers/mixin')
+const Connectable = require('models/Connectable')
+const deleteAllUserRelatedData = require('tasks/users/deleteAllUserRelatedData')
+const CognitoAPI = require('cognito/CognitoAPI')
+
+class DemoUser {
+
+  static async create(timestamp) {
```
DanielSeehausen, line 9: timestamp should likely be auto provided by the create method (see `User.create`).

lib/models/DemoUser/DemoUser.js
```diff
@@ -0,0 +1,45 @@
...
+    const success = insertedCount === 1
+    if (success) delete user._id
+    return user
+  }
+
+  static async findUnActiveDemoUsers(timestamp) {
+    return DemoUser.usersDB.find({ createdAt: { $lt: timestamp } }).toArray()
+  }
+
+  static async removeUnActiveUsers(timestamp) {
```
DanielSeehausen, line 32: to make this a little more clear with an argument, there are two ways we could potentially refine this:
a. make a static getter to define what duration in ms is considered a 'stale/inactive' user and reference it in this method
b. edit the method name to communicate the use case, potentially something like `removeUsersInactiveForMs`

test/lib/models/DemoUser/DemoUser.test.js
```diff
@@ -0,0 +1,59 @@
...
+
+  })
+
+  describe('when demo user in database for more than one month', () => {
```
DanielSeehausen, line 40: this test might be a little misleading with the description and what it tests for

test/lib/models/DemoUser/DemoUser.test.js
```diff
@@ -0,0 +1,59 @@
...
+    it('should remove demo user from cognito pool', async () => {
+      await DemoUser.create(1000)
```
DanielSeehausen, line 42: requested above that `create` extend and delegate to `User.create`, which means the create time argument will unfortunately be lost. The `User.update` method could be used instead to force a creation date in the past.

lib/models/DemoUser/DemoUser.js
```diff
@@ -0,0 +1,45 @@
...
+      id: uuidv4(),
+      role: 'demo',
+      userName,
+      email: `${userName}@ospin.de`,
+      processes: [],
+      devices: [],
+    }
+    await CognitoAPI.createUser(user.userName, password)
+    user.createdAt = timestamp
+    const { insertedCount } = await DemoUser.usersDB.insertOne(user)
```
DanielSeehausen, line 22: one benefit of extending the `User` class is to ensure data gets pushed through the user schema before it is persisted

services/user/src/user/create.js
```diff
@@ -19,9 +19,10 @@ async function create(event, context) {
     email: event.request.userAttributes.email,
     phoneNumber: event.request.userAttributes.phoneNumber,
     userName: event.userName,
-    role: 'user', // currently restricted to just 'user'
+    role: event.role || 'user', // currently restricted to just 'user'
```
DanielSeehausen, line 22: seeing as the new endpoint `createDemoUser` is here now, is this change still required?

lib/models/User/User.js
```diff
@@ -53,9 +53,7 @@ class User {
   }
 
   static async find(userId) {
-    const user = await User.usersDB.findOne({ id: userId })
-    if (user) delete user._id
-
+    const user = await User.findBy({ id: userId })
```
vdiezel, line 56: `findById` ?

services/user/src/user/jobs/removeUnActiveDemoUsers.js
```diff
@@ -0,0 +1,11 @@
+const connectAllModels = require('models/connectAllModels')
+const DemoUser = require('models/DemoUser/DemoUser')
+
+async function removeUnActiveUsers() {
+  await connectAllModels()
+  const now = new Date()
+  now.setMonth(now.getMonth() - 1)
```
vdiezel, line 7: very cool syntax

DanielSeehausen:
closes #437 

big update - probably easiest to review anew!

services/user/test/user/jobs/removeInactiveDemoUsers.test.js
```diff
@@ -0,0 +1,98 @@
...
+      staleUpdatedAndCreatedDemoUserA,
+      staleUpdatedAndCreatedDemoUserB,
+    } = await getVarietyOfUsers()
+
+    const { deletedIds } = await removeInactiveDemoUsers()
+
+    expect(deleteAllUserRelatedData).toHaveBeenCalledTimes(2)
+    expect(deleteAllUserRelatedData).toHaveBeenNthCalledWith(1, staleUpdatedAndCreatedDemoUserA.id)
+    expect(deleteAllUserRelatedData).toHaveBeenNthCalledWith(2, staleUpdatedAndCreatedDemoUserB.id)
+
```
vdiezel, line 92: an assertion for the call to the CognitoAPI would  be nice

services/user/src/user/jobs/removeInactiveDemoUsers.js
```diff
@@ -0,0 +1,31 @@
+const connectAllModels = require('models/connectAllModels')
+const deleteAllUserRelatedData = require('tasks/users/deleteAllUserRelatedData')
+const DemoUser = require('models/DemoUser/DemoUser')
```
vdiezel, line 3: there is no DemoUser directory in models - its in User in my case

services/user/src/user/jobs/removeInactiveDemoUsers.js
```diff
@@ -1,6 +1,6 @@
 const connectAllModels = require('models/connectAllModels')
 const deleteAllUserRelatedData = require('tasks/users/deleteAllUserRelatedData')
-const DemoUser = require('models/DemoUser/DemoUser')
+const DemoUser = require('models/User/DemoUser')
```
ConorTighe1995, line 3: Nice one viktor was about to push this

PR #476, opened 2020-05-26
434 deep clone device

### LINKED ISSUE #434: H3: Clone a virtual device for each new demo account
### Description
When a demo user is created they will automatically be an admin on a virtual device

---

### Proposed Solution(s)
Use the device model to clone a new virtual device based off a pre-existing device in the database. This device will always be online and needs to be added to the dev team accounts for testing. `isVirtual` should be true, this flag will be later used by the front end to stop breaking requested made to the backend by intercepting the request if this is true.

BLOCKED BY: #433 
---

### Deliverable(s)
- [ ] Create an ideal virtual device (something with all the cool streams and non of the client specific ones like celltainer's bag stuff. It should be marked as virtual and as `online` and `idle`)
  **Note:** we will also want to grab a nice finished process (with clean and good looking data in the datagraphs) to use as a sample process on the demo device. Because of this, consider cloning a device (maybe celltainer/FU/w/e), and getting rid of those streams we don't want'/wont look the best. We can then selectively remove those streams we don't want since this is all virtual.
- [ ] On user demo account creation add that account to the device
- [ ] The account should be a **operator** on the device.

Note: these devices are USER specific. the ideal device is cloned so that one demo device is not shared among users!

tests!

### LINKED ISSUE #435: H4: Create a virtual process for each new virtual device.
### Description
When the virtual demo device is created a virtual demo process should be created and added to the virtual demo device

---

### Proposed Solution(s)
Right after the virtual device is cloned when creating the demo account, also clone the demo devices processes to the new user specific cloned virtual device. 

Make it DRY!

BLOCKED BY: #434 
---

### Deliverable(s)
- [ ] Make an ideal sample process for the demo device and add it to it
- [ ] Ensure the processes' datapoints are also associated with the demo process
     ( note: we may have a need to make a `copyProcessWithDataPoints` or something similar)
- [ ] When a virtual device is online the device can be clicked on and the process viewer will show a virtual process.
- [ ] pre-populated virtual process with data when clicked on in reactor.

DanielSeehausen:
closes #434 
closes #435


lib/tasks/devices/createAndAssignDemoDevice.js
```diff
@@ -0,0 +1,41 @@
+const Device = require('models/Device/Device')
+const DemoDevice = require('models/Device/DemoDevice')
+const DeviceAccess = require('ResourceAccess/DeviceAccess')
+const ProcessAccess = require('ResourceAccess/ProcessAccess')
+const Process = require('models/Process/Process')
+const deepCloneDevice = require('tasks/devices/deepCloneDevice')
+
+
+const mapProcessToUser = async (processId, userId) => {
```
vdiezel, line 9: you had to rename it :P - because it doesn't only map to the user but also wipes the clone history and sets it as virtual - which are demo process specific parts - I named it differently.

test/tasks/createAndAssignDemoDevice.test.js
```diff
@@ -0,0 +1,97 @@
...
+    const seededDemoDevice = await Device.find(demoDevice.id)
+    expect(processes).not.toEqual(seededDemoDevice.processes)
+
+    const executableClonedProcess = (await Process
+      .findByQuery({ isVirtual: true, startedBy: null }))[0]
+
+    const { ownerId, access, createdBy } = executableClonedProcess
+    expect(ownerId).toBe(demoUser.id)
+    expect(createdBy).toBe(demoUser.id)
+    expect(ownerId).toBe(demoUser.id)
```
vdiezel, line 76: duplication of line 74

lib/models/Process/Process.js
```diff
@@ -93,6 +95,39 @@ class Process {
...
+
+    process.cloneHistory = [ ...process.cloneHistory, { id: processId, name: process.name } ]
+    const deletableKeys = [ '_id', 'id' ]
+    const creatableProcess = exclude(process, deletableKeys)
+    const newProcess = await Process.create(creatableProcess)
+
+    if (cloneData) {
+      const chunks = await Process.getChunks(processId)
+      await Promise.all(chunks.map(aChunk => {
+        delete aChunk._id
```
ConorTighe1995, line 123: could just call this `chunk`

lib/test-helpers/seeder/Seeder.js
```diff
@@ -74,7 +72,7 @@ class Seeder {
     const newProcessData = generateProcess(processUser, device, processData, processTraits)
     const persistedProcess = await Process.create(newProcessData)
 
-    user && user.processes.push(persistedProcess.id)
+    //if (user) await User.addProcess(user.id, persistedProcess.id)
```
ConorTighe1995, line 75: Was this commented out code left in by mistake?

lib/models/Device/Device.js
```diff
@@ -76,6 +79,16 @@ class Device {
     return device
   }
 
+  static async clone(deviceId, newName) {
+    const device = await Device.find(deviceId)
+
+    const deviceData = exclude(device, [
+      'processes', 'thingName', 'thingId', 'thingArn', 'certificateArn',
```
ConorTighe1995, line 86: Are these placeholders for vars?

lib/test-helpers/seeder/Seeder.js
DanielSeehausen: that was a mistake good 👁️ 
the `Process.create` method takes care of that

lib/models/Device/Device.js
DanielSeehausen: these strings reference keys on the device object
`exclude` will create a new object from device here, without those keys in the array!

lib/tasks/devices/createAndAssignDemoDevice.js
DanielSeehausen: good :eye:
That wasn't intentional from me - I just got mixed up while refactoring the 3 methods to the 2

I changed it back `mapProcessToUserDemoProcess`!

lib/models/Process/Process.js
vdiezel: Often I had a eslint complaining about same variable names in different scopes, usually when I had a temporary variable in a `map`, `reduce` and `filter` call. So I started following the Martin Fowler Pattern here I've seen in his book on refactoring...

lib/test-helpers/seeder/Seeder.js
vdiezel: I think I might have left it there, sorry!

lib/models/Process/Process.js
DanielSeehausen: i need to finish reading that book

test/tasks/createAndAssignDemoDevice.test.js
```diff
@@ -0,0 +1,96 @@
...
+    const originalChunks = await Process.getChunks(startedProcess.id)
+    const clonedChunks = await Process.getChunks(startedClonedProcess.id)
+
+    expect(clonedChunks.every(aChunk => aChunk.processId === startedClonedProcess.id))
+      .toBe(true)
+
+    originalChunks.forEach(deleteChunksIds)
+    clonedChunks.forEach(deleteChunksIds)
+
+    const sortChunks = (chunkA, chunkB) => (chunkA.n - chunkB.n)
```
vdiezel, line 93: I see! The sorting was flickering - can you sort by 'startTime' instead? The `n` property will be deprecated once I find the time to do that.

lib/models/Device/Device.js
ConorTighe1995: Ok I was just wondering why there are vars called 'thingName' etc in the first place that needs to be excluded, assume this is to differentiate from normally devices in the db?

test/tasks/createAndAssignDemoDevice.test.js
DanielSeehausen: will do
some test flickered here I think as this was passing locally!

PR #479, opened 2020-06-02
adds the virtual device and its processes to the created demo user

### LINKED ISSUE #478: The topic for commands is wrong in the documentation
Bildschirmfoto von 2020-05-29 17-33-21

instead of `command` in the Command topics, it should be `commands`.
The `command` key in the payload remains the same.

vdiezel:
finalizes demo user creation!

---
penultimate commit just fixes an error in the docs  - closes #478 

---
last commit adds missing env vars to removeInactiveUsers


lib/models/Device/Device.js
```diff
@@ -25,6 +25,7 @@ class Device {
     return Joi.object({
       id: Joi.string().required().pattern(UUIDV4_REGEX),
       createdAt: unixTimeDefaultNull(),
+      updatedAt: unixTimeDefaultNull(),
```
vdiezel, line 28: cloning of the template device failed because this was missing. We have some devices which do have it (mainly older ones) and new ones that don't. I'd like to keep this field actually.

lib/models/User/User.js
```diff
@@ -18,9 +18,9 @@ class User {
       userName: Joi.string().required(),
       email: Joi.string().email().default(null).allow(null),
       role: Joi.string().valid(...User.USER_ROLES).default('user'),
-      fullName: Joi.string(),
-      phoneNumber: Joi.string(),
-      organization: Joi.string(),
+      fullName: Joi.string().allow(null),
+      phoneNumber: Joi.string().allow(null),
+      organization: Joi.string().allow(null),
```
vdiezel, line 23: `User.update` needed those.

lib/tasks/devices/createAndAssignDemoDevice.js
```diff
@@ -22,20 +23,32 @@ const mapProcessToUserDemoProcess = async (processId, userId) => {
...
+      name: 'Ospin Demo Reactor',
+      access: populatedAccess,
+    })
+
+    const deviceProcesses = await Process.findByQuery({ deviceId: clonedDevice.id })
+    const deviceProcessIds = deviceProcesses.map(({ id }) => id)
+
+    await Promise.all(deviceProcessIds.map(
+      processId => mapProcessToUserDemoProcess(processId, userId),
+    ))
```
vdiezel, line 43: I think we plan to get rid of `processes` in the device as processes already point to their device via the `deviceId` prop. So I changed the way to fetch the associated device processes. Also for the device template we forgot to add the processId to the processes - because this is easy to forget. We shoul clean up the associations very soon

lib/tasks/devices/createAndAssignDemoDevice.js
```diff
@@ -22,20 +23,32 @@ const mapProcessToUserDemoProcess = async (processId, userId) => {
...
+
+    await User.update(userId, {
+      devices: [ clonedDevice.id ],
+      processes: deviceProcessIds,
+    })
```
vdiezel, line 48: Forgot that, too.

services/user/src/user/create.js
```diff
@@ -26,10 +27,15 @@ async function create(event, context) {
 
   await connectAllModels()
 
-  const UserType = newUserData.role === 'demo' ? DemoUser : User
-  await UserType.create(newUserData)
+  if (newUserData.role === 'demo') {
+    const demoUser = await DemoUser.create(newUserData)
+    await createAndAssignDemoDevice(demoUser.id)
```
vdiezel, line 32: I didn't want to put line 32 into `DemoUser.create`. Felt like this violates SoC.

services/user/src/user/createDemoUser.js
```diff
@@ -9,20 +9,11 @@ async function createDemoUser() {
...
-    'demo',
-  )
-
-  const user = await DemoUser.find(UserSub)
+  await CognitoAPI.signUpUser(userName, password, userName + DemoUser.GUEST_EMAIL_DOMAIN, 'demo')
 
   await CognitoAPI.confirmUser(userName)
 
-  return success({ ...user, password: DemoUser.GUEST_PASSWORD })
+  return success({ userName, password })
```
vdiezel, line 16: we need only userName here. Once logged in, a `GET` will fetch the user data anyway.

services/user/test/user/createDemoUser.test.js
```diff
@@ -47,26 +47,15 @@ describe('DEMO USER', () => {
...
-        role: 'demo',
-      }
-
       CognitoAPI.signUpUser = jest.fn(() => ({ UserSub: '53aa617d-452d-466e-8328-a3b1b8d38907' }))
       CognitoAPI.confirmUser = jest.fn()
-      DemoUser.find = jest.fn(() => (userDemoUserData))
 
       const { res: { statusCode }, body: returnedUser } = await fireCreateDemoUserEvent()
-      const dbDemoUser = await DemoUser.find(returnedUser.id)
+      const expectedUsernamePattern = /demo-user-[0-9a-f]{8}/
```
vdiezel, line 54: `createDemoUser` doesn't actually create the user anymore because this is now handled via the post confirmation handler `createUser`.

lib/models/User/User.js
vdiezel: The reason is an inconsistency when creating users: Because AWS sets them to null, they are not considered als `undefined` by Joi, so they are actually set to null on the created user (instead of omitting them) and persisted.

DanielSeehausen: nice good catch - should we also default to null in this case? Think that might be a good idea in general as it will keep our records the same in terms of fields.

lib/tasks/devices/createAndAssignDemoDevice.js
DanielSeehausen: 👍 
+ to getting rid of this

I have been meaning to read up on mongo best practices for joins

lib/tasks/devices/createAndAssignDemoDevice.js
DanielSeehausen: it might be prudent to go through the interfaces here. We have both: `User.addProcess` and `User.addDevice` already!

DanielSeehausen: the `addProcess` could also go either in the `mapProcessToUserDemoProcess`, or in the `Promise.all` which calls it, if you wanted to avoid another `Promise.all` for those two interface methods

lib/tasks/devices/deepCloneDevice.js
```diff
@@ -2,12 +2,14 @@ const Device = require('models/Device/Device')
...
-  const { processes: deviceProcesses, name } = await Device.find(deviceId)
+  const { name } = await Device.find(deviceId)
+
+  const deviceProcesses = await Process.findByQuery({ deviceId })
+  const deviceProcessIds = deviceProcesses.map(({ id }) => id)
 
   const { id: clonedDeviceId } = await Device.clone(deviceId, name)
 
-  await Promise.all(deviceProcesses.map(async processId => {
+  await Promise.all(deviceProcessIds.map(async processId => {
```
DanielSeehausen, line 12: nice update - smart thinking ahead

services/user/src/user/create.js
DanielSeehausen: 💯  solidarity

services/user/test/user/createDemoUser.test.js
DanielSeehausen: 👍 
the test description may need an update now that the object match has been removed!

services/user/src/user/createDemoUser.js
ConorTighe1995: I changed this back to user as Friday before I ran out the door because:

1. I wanted to see if it was related to failing test.
2. Daniel mentions we had a sanity check for user in my PR whick was helpful

Personally im fine with keeping it this way though.

services/user/src/user/create.js
ConorTighe1995: What do you mean by SoC?

lib/tasks/devices/createAndAssignDemoDevice.js
```diff
@@ -22,20 +23,32 @@ const mapProcessToUserDemoProcess = async (processId, userId) => {
...
+
+    return Device.find(clonedDevice.id)
+  } catch (e) {
+    console.log(e)
```
ConorTighe1995, line 52: Should this log stay?

services/user/src/user/create.js
vdiezel: Seperation of concerns

lib/tasks/devices/createAndAssignDemoDevice.js
vdiezel: UH! good catch!

lib/models/User/User.js
vdiezel: Because this one is always fired via Cognito, they are already defaulting to null - but putting it into the schema is probably more explicit and helps documenting,. Will do!

vdiezel:
Guys, I need this to be merged - I'll create another PR with small fixes and will address your requests

services/user/src/user/create.js
ConorTighe1995: good call

services/user/src/user/createDemoUser.js
vdiezel: totally game for the sanity check but it should probably go into the createUser test now

services/user/test/user/createDemoUser.test.js
vdiezel: done!

services/user/src/user/createDemoUser.js
vdiezel: added a new test into createUser

PR #480, opened 2020-06-02
312 reduce the amount of user data returned with devices.

### LINKED ISSUE #312: the device users endpoint returns too much information
when fetching the users of a device we return a lot of information that is not necessary for app operation and also poses a security risk (phone number, email,  all devices the user has access too...). 

Spike into what is really necessary and then reduce the amount of data tha we ship to the frontend. 
Bildschirmfoto von 2020-03-23 08-00-25

ConorTighe1995:
This will change the users data returned with a device by excluding unneeded or sensitive information. As of this pr devices will only return the users id, username and role.

closes #312


services/device/src/serializers/serializeDeviceUsers.js
```diff
@@ -0,0 +1,17 @@
...
+  'notifications',
+  '_id',
+  'email',
+  'phoneNumber',
+  'fullName',
+  'devices',
+  'createdAt',
+  'organization',
+  'processes',
+]
```
vdiezel, line 13: I checked REACTor and I found that we arte using the email of a user, too: 

https://github.com/ospin-web-dev/REACTor/blob/master/src/components/device/manageaccess/UserTable.jsx#L37

I don't know why the heck I used the email there for user comparison instead of ID. The correct way to do this would be to fix the code on REACTor - then we can actually merge this PR.

services/device/src/device/getUsers.js
```diff
@@ -4,20 +4,30 @@ const validateGetUsers = require('../joi/device').getUsers
...
-  const result = await device.getUsers(requestData)
-  return success(result)
+  const users = await device.getUsers(requestData)
+
+  const serializedUsers = []
+
+  users.map(aUser => {
+    const serializedUser = serializeDeviceUsers(aUser)
+    return serializedUsers.push(serializedUser)
+  })
```
vdiezel, line 28:
```js
const serializedUsers = users.map(serializeDeviceUsers)
```

services/device/test/device/getUsers.test.js
```diff
@@ -48,6 +48,7 @@ describe('Device Class', () => {
 
   async function seedUser(userData = {}, traits = []) {
     const newUser = generateUser(userData, traits)
+    serializeDeviceUser(newUser)
```
DanielSeehausen, line 51: as `serializeDeviceUser` is (or at should be if it is not) 🔥  reFeRENtialLY transPAREnt 🔥 
this line may not be doing anything

All the same, I wouldn't get rid of the `require` just yet - we can use it below (see comment)

services/device/test/device/getUsers.test.js
```diff
@@ -100,7 +101,18 @@ describe('Device Class', () => {
...
+              'phoneNumber',
+              'fullName',
+              'devices',
+              'createdAt',
+              'organization',
+              'processes',
+            ]
+
             keysToRemove.forEach(key => delete expected[key])
             expect(body[0]).toMatchObject(expected)
```
DanielSeehausen, line 117:
```suggestion
+           const expectedUser = { ...user, accessGroup: requestersGroup }
-            const keysToRemove = [
-              '_id',
-              'notifications',
-              'email',
-              'phoneNumber',
-              'fullName',
-              'devices',
-              'createdAt',
-              'organization',
-              'processes',
-            ]
-
-            keysToRemove.forEach(key => delete expected[key])
+           expect(body[0]).toMatchObject(serializeUser(expected))
```

DanielSeehausen: hmm that didn't work as I was hoping

DanielSeehausen, line 117: you could use the serialize user in the expected `serializeUser(expected)` and remove the keysToRemove all together

vdiezel: but this way you also implictly test if the serializer is working as expected, so I would keep it OR add seperate tests for the serializer

services/device/src/serializers/serializeDeviceUsers.js
ConorTighe1995: Ive made the following pr and will sort this out today:
https://github.com/ospin-web-dev/REACTor/issues/787

services/device/test/device/getUsers.test.js
ConorTighe1995: I was going to do this initially but thought maybe the reason keys to remove was there was to add confidence to the test?

Your right now that I think of it tho

ConorTighe1995: @vdiezel me and @DanielSeehausen  had a similar discussion already yesterday about ease of change vs confidence.

I could add another test to check for the sensitive data, the pro of that is a descriptive test that brings confidence but the con is another place to change.

services/device/test/device/getUsers.test.js
ConorTighe1995: I made that immutable

services/device/test/device/getUsers.test.js
DanielSeehausen: its a good thing to deliberate on I think. You do make the test less coupled to implementation, and that is always a positive. I'd point out the inverse, that the test becomes more brittle.

That being said, since this is an end to end test, I can equally get behind asserting the response structure independent of a helper method that is used in the inner workings.

services/device/src/device/getUsers.js
```diff
@@ -4,20 +4,26 @@ const validateGetUsers = require('../joi/device').getUsers
...
   const db = await mongo.getDB()
   const device = new Device(db)
 
-  const result = await device.getUsers(requestData)
-  return success(result)
+  const users = await device.getUsers(requestData)
+
+  const serializedUsers = users.map(serializeDeviceUsers)
+
+  console.log(serializedUsers)
```
DanielSeehausen, line 25: this might have snuck by

ConorTighe1995:
> looks good
> re: the test point @vdiezel brought up, I think his point is a good counter to what we talked about while pairing yesterday @ConorTighe1995. I am  with whatever direction that goes in. I suppose in the absence of a serializer test, using the serializer again the in test increases exposure there at the value of a less brittle test.

Ill go with my gut and keep the hardcoded keys in there for the same reason as yesterday, increases the chances of catching a issue and I suppose thats the main goal here.

PR #485, opened 2020-06-03
Small fixes to demo feature

### LINKED ISSUE #487: Small fixes to demo user feature
there are some small fixes required to the demo feature:

- removeInactiveUsers doesn't export the handler in the desired way
- demo suer fetures should be deactivated on environments different from 'demo'

vdiezel:
a series of small fixes / guards / added tests for the demo feature. 

closes #487 

Tested and it worked FIRST TRY


services/user/src/user/create.js
```diff
@@ -20,14 +20,13 @@ async function create(event, context) {
...
     userName: event.userName,
-    role: event.request.userAttributes['custom:role'] || 'user',
   }
 
   validateCreate(newUserData)
 
   await connectAllModels()
 
-  if (newUserData.role === 'demo') {
+  if (global.process.env.STAGE === 'demo') {
```
vdiezel, line 29: With different environments we don't need to pass the role anymore. I prefer this because setting the role required manual action on the Cognito User pool to accept 'custom:role' - this way everything remains documented in the code.

lib/tasks/devices/createAndAssignDemoDevice.js
```diff
@@ -24,32 +24,28 @@ const mapProcessToUserDemoProcess = async (processId, userId) => {
...
+    processId => mapProcessToUserDemoProcess(processId, userId),
+  ))
+
+  await User.update(userId, {
+    devices: [ clonedDevice.id ],
+    processes: deviceProcessIds,
+  })
+
+  return Device.find(clonedDevice.id)
 }
```
vdiezel: jsut removed the try and catch block

services/user/src/user/createDemoUser.js
```diff
@@ -3,8 +3,11 @@ const DemoUser = require('models/User/DemoUser')
 const HTTPDefaultMiddleware = require('middleware/HTTPDefaultMiddleware')
 const { success } = require('http-responses')
 const CognitoAPI = require('cognito/CognitoAPI')
+const { assert, isStage } = require('validation-lib')
 
 async function createDemoUser() {
+
+  assert(() => isStage('demo'))
```
vdiezel, line 10: We don't want to allow calling this endpoint on environments other from 'demo' - because it would break it. Conditional deployment would be even nicer but this does the job for now.

services/user/src/user/jobs/removeInactiveDemoUsers.js
```diff
@@ -4,6 +4,11 @@ const DemoUser = require('models/User/DemoUser')
 const { TimeUtils } = require('utils')
 
 exports.main = async () => {
+
+  if (global.process.env.STAGE !== 'demo') {
+    return
+  }
```
vdiezel, line 10: same here - but we don't want to through an error here, so it just goes by silently.

lib/tasks/devices/createAndAssignDemoDevice.js
```diff
@@ -38,14 +40,11 @@ module.exports = async userId => {
...
-  ))
+  await Promise.all(
+    deviceProcessIds.map(processId => mapProcessToUserDemoProcess(processId, userId)),
+  )
 
-  await User.update(userId, {
-    devices: [ clonedDevice.id ],
-    processes: deviceProcessIds,
-  })
+  await User.addDevice(userId, clonedDevice.id)
```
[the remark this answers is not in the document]
ConorTighe1995: If your worried about writes to the database Mongodb uses a lock system to avoid scenarios like deadlock:

https://stackoverflow.com/questions/6997835/how-does-mongodb-deal-with-concurrent-updates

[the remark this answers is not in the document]
vdiezel: ncie to know, thanks!

services/user/src/user/create.js
DanielSeehausen: potential use case here for a wrapper, like `assert` which coerces the return validators `{ invalid }` to a boolean. That was the original intent of `validate` in that module if you remember, but we were right not to keep it in I think if this is the first use case of it 8 months later

services/user/test/user/create.test.js
```diff
@@ -114,5 +114,35 @@ describe('User Class', () => {
       expect(allUsers.length).toBe(1)
       expect(allUsers[0]).toMatchObject(expectedUser)
     })
+
+    test('should create a demo user in the database when valid data is present and the env is demo', async () => {
+      const userData = {
+        userName: 'Biff',
+        fullName: 'Biff Johnson',
+        email: 'email@thyme.com',
+        phoneNumber: 'boop',
```
DanielSeehausen, line 123: surprised this passed

services/user/test/user/createDemoUser.test.js
```diff
@@ -57,5 +59,14 @@ describe('DEMO USER', () => {
       expect(returnedUser.password).toEqual(DemoUser.GUEST_PASSWORD)
       expect(returnedUser.userName).toEqual(expect.stringMatching(expectedUsernamePattern))
     })
+
+    it('should return 400 when not on demo stage', async () => {
```
DanielSeehausen, line 63: noice

services/user/test/user/create.test.js
vdiezel: email and userName are passed to createUser via Cognito. So it doesn't overwrite them there when creating a demo user

Issue #488, opened 2020-06-04
Deprecate the `processes` property in the device 


vdiezel:
blocked by: https://github.com/ospin-web-dev/REACTor/issues/748

The `processes` array refers to processes associated with the device. Because the processes also already point to their device this reference is not needed anymore. 

- [ ] remove it from the device model
- [ ] remove it from the devices in the database
- [x] fix all tests and other references within the code

Issue #472, opened 2020-05-25
remove 'groupNameAlias' property from the device model


vdiezel:
the property `groupNameAlias` on the device is currently not used. Remove it from the model and update the devices in the database.

afterwards, remove from devices in all environments!

There may be the skeleton of scripts to already do this.

PR #489, opened 2020-06-04
preparing deprecation of processes field on the device

vdiezel:
1. part of #488 and #472 
blocked by:  https://github.com/ospin-web-dev/REACTor/pull/793

Now that our udpate calls are also going through the schema, our migration strategy needs to be slightly more elaborate: 

This PR doesn' remove the property from the schema, as it would break the code (updates on devices wouldn't work anymore). Instead we do this:

Once this is merged, I will migrate the DB. 
Once the migration is done, the next PR on hambda can actually remove the property from the schema. 

Because `groupNameAlias` also has to go through the same flow, so I grabbed it in this PR, too.

## Update
removed missed references as pointed out by daniel and some I found by parsing the whole repo for 'processes'


lib/models/Device/Device.js
```diff
@@ -34,7 +34,7 @@ class Device {
       ownerId: Joi.string().pattern(UUIDV4_REGEX).default(null).allow(null),
       hardwareVersion: Joi.number().integer().default(0),
       processes: Joi.array(), // TODO remove once the property was removed in the DB
-      groupNameAlias: Joi.array().default([]),
+      groupNameAlias: Joi.array(),
```
ConorTighe1995, line 37: Should this not be totally removed if your deprecating groupNameAlias? Just making sure you didn't leave it in my mistake.

vdiezel: I can't remove it in one go because our `update` device calls also go through the schema. When I remove it and the PR gets merged, until I have updated the database, all calls for update device will fail due to failed validation. If I migrate the DB first, then all calls to device update will restore `groupNameAlias`. So instead this will go in multiple steps.

vdiezel:
> Because this stuff could potentially sneak by when destructuring, I did two searches:
> `Ag -Q '{ processes'` and `Ag -Q 'processes }'`
> But that's far from complete - I am unsure how to look for it everywhere outside of examining every `processes` reference :/ or looking up some multiline parser method to find `processes` between brackets
> 
> These might have been left in intentionally, but if not:
> 
> [addProcess](https://github.com/ospin-web-dev/hambda/blob/488-deprecate-processes-on-device/lib/models/Device/Device.js#L141-L146)
> 
> [removeProcess](https://github.com/ospin-web-dev/hambda/blob/488-deprecate-processes-on-device/lib/models/Device/Device.js#L148-L153)
> 
> [removeProcessFromAllDevices](https://github.com/ospin-web-dev/hambda/blob/488-deprecate-processes-on-device/lib/models/Device/Processes/DeviceProcesses.js#L3-L8)
> 
> [addProcessesToDevice](https://github.com/ospin-web-dev/hambda/blob/488-deprecate-processes-on-device/services/process/test/process/utils/addProcessesToDevice.js#L3)
> This file can be deleted - I don't see it being used anywhere. I missed this reference when refactoring the model.
> 
> [noticed this as well. Unrelated to this PR, but I don't see it being used anywhere and it is now deprecated by the interfaces. Believe its safe to delete if you wanted to sneak in with this PR](https://github.com/ospin-web-dev/hambda/blob/488-deprecate-processes-on-device/services/process/test/process/utils/addProcessesToUser.js)

Man, this was some sloppy work for me. Thanks for taking the effort to point me to the correct direction!

lib/models/Device/Device.js
```diff
@@ -2,7 +2,6 @@ const uuidv4 = require('uuid').v4
 const { UUIDV4_REGEX, exclude } = require('utils')
 const Joi = require('@hapi/joi')
 const { unixTimeDefaultNull } = require('joiUtils')
-const DeviceProcesses = require('models/Device/Processes/DeviceProcesses')
```
DanielSeehausen, line 5: v nice

PR #491, opened 2020-06-05
490 virtual processes

### LINKED ISSUE #490: Create Process flow needs a way to be `isVirtual === true`
### Description
When creating a process from the demo branch, the process is not flagged as `isVirtual === true`

There are a few ways we could tackle this, one way being: updating the endpoint to accept that data

A potentially even easier option would be to update `services/process/src/process/create.js` to check if the device (that is has already looked up) is virtual. We could simply pass along the device's `isVirtual` to `Process.create`

```js
async function create(event) {
  const requestData = {
    deviceId: event.queryStringParameters.deviceId,
    data: {
      name: event.body.name,
      deviceId: event.queryStringParameters.deviceId, // TODO: should this be a path parameter, understanding device is required (i.e. is this really create device process)?
      createdBy: event.requestingUserId,
      description: event.body.description,
      deviceDescription: event.body.deviceDescription,
      entryPhaseId: event.body.entryPhaseId,
      cloneHistory: event.body.cloneHistory,
      comment: event.body.comment,
      eventTriggers: event.body.eventTriggers,
    }
  }

  validateCreate(requestData)

  await connectAllModels()

  const device = await Device.find(requestData.deviceId)

  assertMany([
    () => canFind(device, `Device ${requestData.deviceId}`),
    () => hasSufficientResourceAccess(device, requestData.requestingUserId, RESOURCE.DEVICE, DEVICE_ACTIONS.OPERATE)
  ])

  const process = await Process.create({ ...requestData.data, isVirtual: device.isVirtual })
  return success(process)
}
```

DanielSeehausen:
closes #490


services/process/src/process/create.js
```diff
@@ -45,7 +45,11 @@ async function create(event) {
     () => hasSufficientResourceAccess(device, requestData.requestingUserId, RESOURCE.DEVICE, DEVICE_ACTIONS.OPERATE)
   ])
 
-  const process = await Process.create(requestData.data)
+  const process = await Process.create({
+    ...requestData.data,
+    isVirtual: device.isVirtual,
```
ConorTighe1995, line 50: Nice so this means we will now be labeling what processes are virtual too?

I could make I ticket once this goes in to create a virtual distinction in the UI.

ConorTighe1995: After thinking about it I guess there is no real point as all processes will be virtual once you click a virtual device.

DanielSeehausen: I'm with you - can leave it for now. something to potentially add in the future.

PR #495, opened 2020-06-09
492 put user endpoint

### LINKED ISSUE #492: Add a PUT /user/:id endpoint
Their is currently no `PUT /user` endpoint for updates of the user in the database

### Deliverables
 - [ ] create test coverage for a PUT `/user` endpoint
  - only admins and the user themselves should be able to change a users data via this endpoint for the time being
 - [ ] pass the tests!
 - [ ] write up a REACTor ticket to provide an interface in the front end (potentially in the `Profile` view) so that users can change this

DanielSeehausen:
closes #492 !

Reviewing commits will separate the very little noise in the file diffs.

The only deletion diff here was all on unrelated lint/copy changes

All of the tools we have made over time for this stuff worked perfectly - 0 refactoring, which was:
image


services/user/test/user/update.test.js
```diff
@@ -0,0 +1,166 @@
...
+        expect(res.statusCode).toBe(422)
+        assertContainsAll(message, [ 'Validation failed', 'email' ])
+
+        const userPost = await User.find(user.id)
+
+        expect(userPost.userName).toEqual(user.userName)
+        expect(userPost.email).toEqual(user.email)
+      })
+
+      test('should return 422: Request data does not pass validation when no editable field is passed to be updated', async () => {
```
vdiezel, line 127: maybe this would be better served with this: https://developer.mozilla.org/de/docs/Web/HTTP/Status/304 ?

lib/models/User/User.js
```diff
@@ -65,6 +65,26 @@ class User {
     return user
   }
 
+  static async update(userId, userData) {
+    const user = await User.find(userId)
+
+    const updatedUser = {
+      ...user,
+      ...userData,
+      updatedAt: new Date().getTime(),
```
vdiezel, line 74: `Date.now()` often better (is a static method, doesn't require to create a new instance

services/user/src/user/update.js
```diff
@@ -0,0 +1,53 @@
...
+    body: { userName, email, fullName, phoneNumber, organization },
+  } = event
+
+  const updateableData = deleteUndefinedProperties({
+    userName,
+    email,
+    fullName,
+    phoneNumber,
+    organization,
+  })
```
vdiezel, line 30: wouldn't it be simpler to defer what the allowed keys are to Joi and simply pass an `update` object to the endpoint with the desired values I'd like to update?

Then the validation in line 32 will decide which properties are allowed (instead of `updatableData`), so per default values that are not explictly passed into the update call will be not present instead of undefined.

If somebody explictly passed `email = undefined`, it will pass the validation on line 32 (because to Joi a value of undefined is the same as the key not being present at all, so it wont validate it) but the user schema will reject it anyway.

services/user/src/joi/user.js
```diff
@@ -19,6 +19,15 @@ const create = createSchemaAssertion(Joi.object({
...
 }))
 
+const update = createSchemaAssertion(
+  Joi.object({
+    userName: Joi.string(),
+    email: Joi.string().email(),
+    fullName: Joi.string(),
+    phoneNumber: Joi.string(),
+    organization: Joi.string(),
+  }).or('userName', 'email', 'fullName', 'phoneNumber', 'organization'),
```
vdiezel, line 29: I can't tell what the `or` is for here - it lists the exact same optional properties again

services/user/src/user/update.js
ConorTighe1995: It looks like its to avoid overwriting the missing values in the PUT? PUT can update specific values in a db object/item. So I if I only pass `id` and the rest are undefined I'm guessing this is to shave off the undefined values so they don't overwrite the vars in the db, only PUTTING in the new `id` alone. We don't want Joi to reject a PUT request or anything to blow up just because values are missing because its ok to send 1 value in a PUT.

Thats how Im interpreting it anyway, and I wonder is there and flags you can pass to the Mongodb document update call to do this which would reduce the code and complexity.

lib/models/User/User.js
DanielSeehausen: 💯 agree. I copied this over and did not pay close enough attention. will update

services/user/src/joi/user.js
DanielSeehausen: i will explain this in the other comment re: the undefined stuff

services/user/test/seeds/event-generators/updateUserEvent.js
```diff
@@ -0,0 +1,17 @@
+const updateUserEvent = (requestedUserId, requestingUserId, body) => ({
+  pathParameters: {
+    userId: requestedUserId,
+  },
+  headers: {
+    'Content-Type': 'application/json',
+    'content-type': 'application/json'
```
ConorTighe1995, line 8: Duplicate headers?

services/user/src/user/update.js
DanielSeehausen: as discussed in meeting - that's the short and skinny of it!

services/user/test/seeds/event-generators/updateUserEvent.js
DanielSeehausen: good eye - this is a casing thing (when we were writing middleware I recall there was inconsistent handling based on casing, so just send both to mimic a client doing either)

PR #496, opened 2020-06-10
Quickfix: adds a default to eventTrigger and a safe guard when the property doe…

vdiezel:
…s not exists on processes yet

Good part: This broke only when thge users didn't have any triggers set. So there was no issue with the fireTriggers endpoints not working.


lib/models/Process/Streams/ProcessStreams.js
```diff
@@ -6,6 +6,9 @@ class ProcessStreams {
 
   static getAllStreamEventTriggers(process) {
     const { eventTriggers } = process
+
+    if (eventTriggers === undefined) return []
+
     const streamIds = Object.keys(eventTriggers)
```
vdiezel, line 12: this broke because our schema for Processes didn't have the eventTriggers set as a default property.

lib/models/Process/Process.js
```diff
@@ -47,7 +47,6 @@ class Process {
   static async create(processData) {
     const {
       createdBy: ownerId,
-      deviceId,
```
vdiezel, line 50: wasn't used anymore.

lib/models/Process/Process.js
```diff
@@ -35,7 +35,7 @@ class Process {
       createdAt: unixTimeDefaultNull(),
       pinned: Joi.boolean().default(false),
       comment: Joi.string().allow('').max(250).strict(),
-      eventTriggers: Joi.object(),
+      eventTriggers: Joi.object().default({}),
```
ConorTighe1995, line 38: Should this default to a empty object or should we be a bit more descriptive with the default?

ConorTighe1995: I guess this is the easiest thing to do while solving the undefined issue below.

vdiezel: the structure of event triggers can be quite different so for now I didn't want to restrict it further.

PR #503, opened 2020-06-16
fix insertMany when no chunks are present

vdiezel:
`insertMany` failed when data was an empty array

fixes this: 

Bildschirmfoto von 2020-06-16 15-26-27


DanielSeehausen:
(clicked update branch)

services/dataPoints/test/data/triggers/evacuateLastChunk.test.js
```diff
@@ -20,7 +20,9 @@ describe('evacuateLastChunk', () => {
   afterEach(async () => {
     jest.clearAllMocks()
     redis.clear()
-    await db.collection('testData').drop()
+    try {
+      await db.collection('testData').drop()
+    } catch (e) {}
```
ConorTighe1995, line 25: This will just fall silent, not a big deal but may this should just log something.

PR #504, opened 2020-06-16
501 support endpoint

### LINKED ISSUE #501: New endpoint for support user details
We need to extend the REST API to provide support users details to both Reactor and the up coming support application, this would return a payload of all users on a device that belong to the group support.

Agree with team on a appropriate url, examples:
`device/:deviceId/support/users`
`device/:deviceId/support`

DOD: I should be able to hit the url with a `deviceId` and get the expected payload of all support uses associated with that `deviceId`.

ConorTighe1995:
closes #501 

We can now have access rights for support users, the serializer on getUsers  will handle the support users by returning the email and phone number for those accounts. Ive also done some refactoring in the PR and updated/added tests to cover support users.


services/device/src/serializers/serializeDeviceUsers.js
```diff
@@ -12,6 +12,19 @@ const EXCLUDED_ATTRIBUTES = [
...
+  '_id',
+  'fullName',
+  'devices',
+  'createdAt',
+  'organization',
+  'processes',
+]
+
+module.exports = user => {
+  if (user.accessGroup === 'support') {
```
ConorTighe1995, line 26: After talking to daniel about an endpoint vs handling it in the serialization code, I decided he was right, this saves us creating a new endpoint and a lot of time.

lib/test-helpers/seeder/generators/device.js
```diff
@@ -55,6 +55,16 @@ const traitDispatch = {
...
+  USER_AS_SUPPORT: (device, user) => {
+    let access = DeviceAccess.create()
+    access = DeviceAccess.addUserToGroup(access, user.id, 'support')
+
+    return {
+      access,
+      ownerId: user.id,
+    }
+  },
+
```
DanielSeehausen, line 67: I don't see this being used anywhere in this diff
you could use it here:
https://github.com/ospin-web-dev/hambda/pull/504/files#diff-e2cddc54d90b387470fd72bd3616aff3R141-R144
where you create the user (and then you won't need to add the user to the group)

Or you could just remove this addition :p, as with the interfaces the traits will/are getting less and less use

ConorTighe1995:
> this looks great. I am blocking but only because there is something I want to make sure we at least discuss before this goes in:
> 
> as this is now, even device admins have the ability to add or remove device support users.
> we might want to disable that and make it so only users of type 'developer' or 'admin' (not on the device, but instead their `role` type) can perform that addition.
> 
> The validation (which would probably happen in `device/src/device/modifyAccess`) might get a little tricky, (and the big device interface PR will change that stuff a lot) so maybe we just go the simplest route: if the user that is modifying access does not have role 'developer/admin', and the new group is 'support', we just throw 403 error right there?
> 
> What do you guys think? Are there any other things we need to consider with a resource user role in the system?

Good foresight Daniel! For now we could do the 403 response, also we could remove all support users from the addition list or deletion list in the web application, that interaction with the server should probable be done on the new support application. We could have users of Reactor submit a request to add a new support user through a form opened by a button on the device table and this request could appear in the support application? let me know what you think because I'm just spit balling ideas.

DanielSeehausen:
> Good foresight Daniel! For now we could do the 403 response, also we could remove all support users from the addition list or deletion list in the web application, that interaction with the server should probably be done on the new support application. We could have users of Reactor submit a request to add a new support user through a form opened by a button on the device table and this request could appear in the support application? let me know what you think because I'm just spit balling ideas.

The 403 return sounds good. Agree 100% that on the webapp a user shouldn't see support as one of the group options when modifying a device users access. 

The request support user might make sense in the future. Let's wait to see how this takes shape - I imagine adding support users will be largely an internally decided/completed thing instead of something individual device owning teams do themselves. In the case of celltainer being a support user, that would be because they are the issuers of the machines, but not the owners themselves, so again probably an internally assigned thing.

services/device/src/serializers/serializeDeviceUsers.js
```diff
@@ -12,6 +12,19 @@ const EXCLUDED_ATTRIBUTES = [
...
 ]
 
-module.exports = user => (
-  exclude(user, EXCLUDED_ATTRIBUTES)
-)
+const EXCLUDED_SUPPORT_ATTRIBUTES = [
+  'notifications',
+  '_id',
+  'devices',
+  'fullName',
```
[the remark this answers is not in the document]
ConorTighe1995: When talking in the other PR remember we talked about fullName vs userename.

Username is more consistent and if we had firstName stored then that would be a better option as opposed to giving away a support users full name.

services/device/src/device/getUsers.js
```diff
@@ -41,6 +40,7 @@ async function getUsers(event) {
 
   const serializedUsers = serializeDeviceUsers(users, device.access)
 
+  console.log(serializedUsers)
```
DanielSeehausen, line 43: this guy snuck in here

ConorTighe1995:
> its looking good from my end! just the console.log that snuck in.

fixed

PR #505, opened 2020-06-16
375 378 device interface

### LINKED ISSUE #375: Transition `device` to a `Device` model
- [ ] Continue that transition to a more functional interface that matches the other models (`Event`, `UIConfig`, etc. should be decent examples). the devices/src/device/device.js file should be removed after this ticket is finished
- [ ] Device should become a `connectable` as part of this transition
- [ ] Update the seeder to make use of `Device.schema` instead of making Devices through a different route
- [ ] click test thoroughly! add any high value tests you see missing!

Update all lambda functions only to call the new model (breaking out authorization into the controllers will be in another ticket)

Note: this is a huge ticket that touches a lot, but it is not a great candidate to get merged at different times. Understanding this, consider making a detailed commit plan.

### LINKED ISSUE #378: Move Device authorization out of base model/interface calls
related to #376
related to #377
blocked by #375 

Right now a lot of our model methods are doing device authorization directly in them.
(Note: this does not include dependency assertion, e.g.: 'a device needs to be found by that id for you to perform this authorization on a device`)

Choose one of:
OR
Give this a read!
One path forward we could consider is to either make a new Class extend the core Device interface (or use connectable on the base Class). This extension Class would have methods that look like what we already have on the Device class (functionality combined with authorization). These methods would continue to take a userId and do the same stuff on the model. This would save us from 'robbing ourselves blind' as Fowler put it.

DanielSeehausen:
closes #375 
closes #378 

ready for review!
highly recommend reading it commit wise - lots in here but the commits are broken up pretty cleanly.
It would have been heavily in the negative just from re-use of methods, if not for the sub interfaces of device which took up a lot of boiler plate code lines :(

There were some test flickers but I can't recreate them :/
I believe they may be from lack of `Promise.all`s in tests... if you see any issues after this gets merged please let me know

sorry for the size! I hope the interfaces prove useful


DanielSeehausen:
the only two code climate things I did not mark as invalid aren't showing up here for some reason

lib/models/Device/Device.js
```diff
@@ -77,6 +77,13 @@ class Device {
     return device
   }
 
+  static async findManyBy(query) {
+    const devices = await Device.devicesDB.find(query).toArray()
+    devices.forEach(device => delete device._id)
```
vdiezel, line 107: would `const devices = await Device.devicesDB.find(query).select({ _id: 0 }).toArray()`
work to get rid of the `_id`

lib/models/User/Notifications/Notification.js
```diff
@@ -30,7 +30,7 @@ class Notification {
         deviceId: Joi.string().pattern(UUIDV4_REGEX), // Only on invitations TODO: this needs to be made a sub schema when device model is compelted
         message: Joi.string().default(''),
       }).required(),
-      read: Joi.bool(), // TODO: is this still required?
+      read: Joi.bool().default('false'), // TODO: is this still required?
```
vdiezel, line 33: it might be very handy for persisted messages: Once a user had a look at them, we could stop showing the bell red (we arent doing that currently, but we probably want at some point)

lib/validation-lib.js
```diff
@@ -125,11 +125,23 @@ exports.userIsResourceOwner = (requestingUserId, resource) => {
...
+  if (collection.includes(value)) {
+    return { invalid: false }
+  }
+
+  return {
+    invalid: true,
+    errorMsg: errorMsg || `${value} was not found in collection ${collection}`,
+    errorFn: throwConflictError,
+  }
+}
```
vdiezel, line 138: 👍

services/device/src/device/triggers/updateState.js
```diff
@@ -18,27 +24,45 @@ const Device = require('../device')
...
-    state: event.state,
-  }
-  if (requestData.processId === 'null') delete requestData.processId
+  let {
+    processId,
+    deviceId,
+    state,
+  } = event
+
+  // TODO: where is this coming from? we have test coverage for it so I didn't want to remove it
```
vdiezel, line 33: This was a workaround for an inconsistency in the payload: For the state idle we received the string "null", instead of the value `null` which confused the endpoint because it thought its a processId. 

I don't know the state of this on the firmware side. 

With your changes to the logic you can change line 51 to just be 
```js
runningProcessId: processId
```

lib/models/connectAllModels.js
```diff
@@ -15,6 +15,7 @@ const CONNECTABLE_MODELS = [
   Process,
   UIConfig,
   DataPoints,
+  DataPoints,
```
vdiezel, line 18: why is this duplicated?

lib/models/Device/Config/DeviceConfig.js
```diff
@@ -26,13 +26,9 @@ class DeviceConfig {
...
   }
 
-  static create() {
-    const newConfig = {
-      global: {},
-      streams: [],
-    }
-
-    return DeviceConfig.schema(newConfig)
+  static new(data = {}) {
```
vdiezel, line 29: why using this a wrapper - you think `schema` is not descrtiptive enough for an exposed API function name?

lib/models/Device/Device.js
```diff
@@ -11,7 +11,7 @@ const mixin = require('functional-helpers/mixin')
 
 class Device {
 
-  // TODO: make extended class for bioreactor
+  // TODO: make extended class(es) when this gets a second device type
   static get DEVICE_TYPES() {
     return new Set([ 'bioreactor' ])
```
ConorTighe1995, line 21: Does this need to be a set if this is hardcoded information?

services/device/src/device/list.js
```diff
@@ -1,23 +1,24 @@
...
+
+  validateList({ requestingUserId })
 
-  validateList(requestData)
+  await connectAllModels()
 
-  const db = await mongo.getDB()
-  const device = new Device(db)
-  const result = await device.list(requestData)
+  // TODO: assert that requesting user is found
```
ConorTighe1995, line 17: Forgot about this?

services/device/src/serializers/serializeDeviceUsers.js
```diff
@@ -12,6 +14,14 @@ const EXCLUDED_ATTRIBUTES = [
...
 
-module.exports = user => (
-  exclude(user, EXCLUDED_ATTRIBUTES)
+// For a given array of user objects
+// -> remove undesired properties
+// -> locate the device group a user belongs to in an access object and add the name to the user object
+module.exports = (users, deviceAccess) => (
+  users.map(user => {
+    const serializedUser = exclude(user, KEYS_TO_REMOVE)
+    serializedUser.accessGroup = DeviceAccess.getUsersGroup(deviceAccess, user.id).name
```
ConorTighe1995, line 23: This is access name as opposed to device name yea?

should name not be role to be consistent?

lib/models/User/Notifications/UserNotifications.js
```diff
@@ -8,6 +8,14 @@ class UserNotifications {
     ))
   }
 
+  // TODO: test coverage - expand this interface (device invitations are a type of notification)
+  static getDeviceInvitationIds(user) {
+    return user.notifications
+      .filter(note => note.topic.split('/')[1] === 'invitation' && note.topic.split('/')[0] === 'device')
```
ConorTighe1995, line 14: Looks like your extracting url words here? Id throw theses in consts to make it easier to read.

services/device/test/device/setStreamDefault.test.js
```diff
@@ -72,6 +72,12 @@ describe('Device Class', () => {
     test('should set a streamDefault value when a streamId that exists on the device is used', async () => {
 
       const streams = Array.from({ length: 2 }, () => generateStream())
+      streams[0].value = true
```
ConorTighe1995, line 75: why is value a boolean here and a int below?

lib/models/Device/Config/DeviceConfig.js
DanielSeehausen: good :eye:
there is no reason this is still here from my end. i will look into replacing this now

DanielSeehausen: updated!

lib/models/Device/Device.js
DanielSeehausen: no it certainly doesn't need to be, but I do my best to preference sets when an array should only contain unique values -- the hope is that its more for code communication than any methods set provides us

lib/models/Device/Device.js
DanielSeehausen: made a big update around this just now - thanks for suggestion

services/device/src/device/triggers/updateState.js
DanielSeehausen: updated

lib/models/connectAllModels.js
DanielSeehausen: Did not even realize doing this. unsure how it happened! good 👁️

lib/models/User/Notifications/Notification.js
DanielSeehausen: nice thanks for explanation

services/device/src/device/list.js
DanielSeehausen: i'd like to leave this TODO here. while this is a poor excuse, it is outside of the scope of this PR and I don't want to muck with it now with so many changes already here. This is because it might have unforeseen consequences on the front end, but these are all authenticated routes so we will want to update it at some point

services/device/src/serializers/serializeDeviceUsers.js
DanielSeehausen: Correct - more specifically this is a device access group's name
role might not work well because we have something called `role` on users, which is separate from the concept of an access group on a resource

`name` is actually something on every access group object, which is what we get out of `getUsersGroup`

I may be misunderstanding what you are saying, though, so please let me know if the above does not suffice as an explanation

services/device/test/device/setStreamDefault.test.js
DanielSeehausen: good question
this is to provide test coverage for streams with different value types
I probably should have broken these out to be more tests with more descriptions to make that obvious

the interface exposed some failing tests (or falsely passing tests?) on corner cases IIRC, so I kinda rushed through this trying to change the minimum possible while providing coverage

lib/models/User/Notifications/UserNotifications.js
DanielSeehausen: i agree - if you will let me I'd like to be lazy and leave this for now. I left the TODO recognizing this interface needs some expansion. I'm hoping next time any of us are working in it we can take the time to do that. For this, I just copied over what was already being done directly in to this file to try to stay on track.

services/device/src/device/list.js
ConorTighe1995: All good with TODOs, just seen asserts at other parts in the pr and thought id shout it out in case there will be a piece of code merged that should have an assert that could later cause issues, but if its just for the future its fine

services/device/src/serializers/serializeDeviceUsers.js
ConorTighe1995: I see I just wanted to make sure it was the name of access group

PR #509, opened 2020-06-22
508 add process description schema

### LINKED ISSUE #508: Provide SCHEMA for process description
# Description

Currently, process description has no schema. In fact, the front end seems to be the consistent source of truth for what a process schema is.

As this may get serialized in the future, and we want to make sure we are sending consistent payloads to the device, add in a SCHEMA for a process description for Process interface.

DanielSeehausen:
closes #508 

adds SCHEMA for process description and makes use of it in the `Process` interface
Left the schema pretty lenient. There are several validations we could do (and will need to do on the front end in the interface helper methods) such as:

- assert values passed for `iterationOf` match a phase key
- assert that if an object has an `iterationOf` that is not null, then its values for `transition`, `duration`, `groupName`, `downstreams`, `name` all match
- assert the numbers that make up the keys for `downstreams` match downstream ids found in the parent Process object's `deviceDescription`(?)! that would be a complex one

If you haven't seen the JOI function `Joi.object().pattern(...)` before, take a look at the last test in the new test file (which I put there mostly to document what is going on). It is new for me, and at a minimum wasn't obvious what was going on by the name alone, so let me summarize the docs:

`pattern`s first arg is for the keys, and the second is for the value of said keys. [The documentation shows some pretty crazy ways it can be expanded](https://hapi.dev/module/joi/api/?v=17.1.1#objectpatternpattern-schema-options).


lib/models/Process/Description/ProcessDescription.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  static get PHASE_SCHEMA() {
+    return Joi.object({
+      name: Joi.string().allow(null).default(null),
+      downstreams: Joi.object().pattern(
+        Joi.number(),
+        ProcessDescription.PHASE_DOWNSTREAM_TARGET,
+      ).required(),
+      transition: Joi.string().valid(...this.TRANSITION_TYPES).default('manual'),
```
vdiezel, line 23: this property is required

DanielSeehausen: i will update

DanielSeehausen:
I will run processes through mongoDB before going forward, per suggestion. good idea

DanielSeehausen:
@vdiezel I updated the `downstreams` object in `ProcessDescription` to allow `eventTriggers`. I imagine that's supposed to be there instead of some of those downstream objects having the key there incorrectly. Please let me know if that's not the case. Are there any other keys that come to mind that might be missing?

Going forward, we should strict about defining data objects like this in the schemas, and fleshing out those that are still just plain `Joi.object()`

scripts/ValidateAgainstSchema.js
```diff
@@ -0,0 +1,46 @@
+const Joi = require('@hapi/joi')
+
+const mixin = require('../lib/functional-helpers/mixin')
+const Connectable = require('../lib/models/Connectable')
+
+
+class ValidateAgainstSchema {
+
+  static async validate(document, schema) {
```
ConorTighe1995, line 9: When you say document you just mean json object?

DanielSeehausen: mongodb document
which is a javascript object, if not quite JSON

DanielSeehausen:
> eventTriggers is not supposed to be on the process description anymore - this was an old implementation that I changed - this should also only be the case on dev (I hope).
> 
> Event triggers now live as a top level key on the process

I will get rid of them then and revert this

PR #513, opened 2020-07-01
upadtes publish command to check the payload size before sending

### LINKED ISSUE #512: put stop gap for payload size in mqtt
fail the lambda and send a loud and clear error message that the message was too big

DanielSeehausen:
Short and sweet!

closes #512 

adds a new handled error using the library for [413](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413#:~:text=The%20HTTP%20413%20Payload%20Too,a%20Retry%2DAfter%20header%20field.)


lib/IotDataAPI.js
```diff
@@ -29,6 +31,8 @@ const RETURN_DISPATCH = {
 
 class IotDataAPI {
 
+  static get PAYLOAD_BYTE_LIMIT() { return 128 * 1000 }
```
ConorTighe1995, line 34: Whats the reason for this limit? should this be put in a const? why not this to save memory:

`static get PAYLOAD_BYTE_LIMIT() { return 128000 }`

That is a bit extreme but its been ingrained into my head to save memory where you can :D

DanielSeehausen: re: the number notation
you have a good point. My instinct is to also avoid the multiplication!

I actually saw this pattern being used with time quite often (instead of showing ms directly, someone would write `1000 * 60 * 30`, for example. This was to communicate to the reader that its 30 minutes, and also make changing the value directly very easy.

I was thinking the same thing here with kBs and did it out of habit, but perhaps it doesn't have the same value as time since its dealing with factors of 10 (while with time, few are masochistic enough to memorize how many minutes 1800000ms is)

I will change it!

Re: putting it in a const: 
I perceive putting it in a getter (`static get ...`) as providing the same functionality as a const. The only difference is, that the Class now owns its limit, which makes sense to me and ensures external callers can reference that limit (like I do in the tests)

